PR #25, opened 2020-07-30
style.css setup

biancamacias:
this is a PR that sets up the style.css file -- there is basic styling on here just for the ```body```


biancamacias:
> LGTM, although it will be phased out/replaced if we decide on using a CSS template.

yeah, thats why i just added like background color and margin. like a default that wont matter once we use a template.

PR #31, opened 2020-07-30
implemented UniversityData.java class

biancamacias:
i used a builder pattern to create a UniversityData.java class. i still need to add ```get``` methods to quickly just get the ```uniName```, ```uniCity```, etc.

### DIFF
--- a/capstone/src/main/java/com/google/univiz/UniversityData.java
+++ b/capstone/src/main/java/com/google/univiz/UniversityData.java
@@ -89,7 +89,22 @@ public final class UniversityData {
     }
 
     public UniversityData build() {
-      return new UniversityData(this);
+      UniversityData uni = new UniversityData();
+      uni.uniName = this.uniName;
+      uni.uniCity = this.uniCity;
+      uni.latitude = this.latitude;
+      uni.longitude = this.longitude;
+      uni.isMainCampus = this.isMainCampus;
+      uni.urbanizationDegree = this.urbanizationDegree;
+      uni.admissionRate = this.admissionRate;
+      uni.avgSAT = this.avgSAT;
+      uni.numOfUndergrads = this.setNumberOfUndergrads;
+      uni.avgCostOfAttendance = this.avgCostOfAttendance;
+      uni.numOfStudentsInDebt = this.numOfStudentsInDebt;
+      uni.carnegieSizeDegree = this.carnegieSizeDegree;
+      uni.numOfUndergradMen = this.numOfUndergradMen;
+      uni.numOfUndergradWomen = this.numOfUndergradWomen;
+      return uni;
     }
   }
 }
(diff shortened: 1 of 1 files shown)


capstone/src/main/java/com/google/univiz/UniversityData.java
```diff
@@ -0,0 +1,110 @@
...
+    private float admissionRate;
+    private float avgSAT;
+    private int numOfUndergrads;
+    private int avgCostOfAttendance;
+    private int numOfStudentsInDebt;
+    private int carnegieSizeDegree;
+    private float numOfUndergradMen;
+    private float numOfUndergradWomen;
+
+    public UniversityBuilder setName(String name) {
```
mmurray22, line 21: can there be one function which is just one massive constructor so you can set all these fields at once; like a `public UniversityBuilder setAll(name, uniCity, ...)`

capstone/src/main/java/com/google/univiz/UniversityData.java
```diff
@@ -0,0 +1,110 @@
...
+    private int isMainCampus;
+    private int urbanizationDegree;
+    private float admissionRate;
+    private float avgSAT;
+    private int numOfUndergrads;
+    private int avgCostOfAttendance;
+    private int numOfStudentsInDebt;
+    private int carnegieSizeDegree;
+    private float numOfUndergradMen;
+    private float numOfUndergradWomen;
```
ihsan314, line 19: Instead of setting the properties of the builder, you can do this instead:
```suggestion
public final class UniversityData {

  private String uniName;
  private String uniCity;
  private float latitude;
  private float longitude;
  private int isMainCampus;
  private int urbanizationDegree;
  private float admissionRate;
  private float avgSAT;
  private int numOfUndergrads;
  private int avgCostOfAttendance;
  private int numOfStudentsInDebt;
  private int carnegieSizeDegree;
  private float numOfUndergradMen;
  private float numOfUndergradWomen;
  public final class UniversityBuilder {
    UniversityData uniData = new UniversityData();
```
and set the properties of `uniData` instead. Then in the build function, you can return `uniData`.

biancamacias: do i keep the ```uniData.uniName = this.uniName``` lines? or do i just do ```return uniData```?

mmurray22: [This](https://www.vogella.com/tutorials/DesignPatternBuilder/article.html) seems like a good ref. Not sure if this is what Ihsan is getting at, but it's one way of doing it.

ihsan314: Just `return uniData`.

biancamacias: ok i will try that.

biancamacias:
still getting these ```cannot find symbol``` errors, just now in the ```UniversityData``` class 
image

capstone/src/main/java/com/google/univiz/UniversityData.java
ihsan314: > still getting these `cannot find symbol` errors, just now in the `UniversityData` class
> image

I forgot to mention, every instance of `this.property` would now be replaced by `uniData.property.`

PR #32, opened 2020-07-30
Collegescorecard interface

mmurray22:
NOTE: Will not pass Java Maven CI test until merged with the universityDataClass branch

This PR sets up the basic interface for the class which will communicate with the API. This interface will be used by all servlets in need of API data! Also, if anyone has extra functions they want to see in this interface, leave that in your comment and I will make sure to add it!


capstone/src/main/java/com/google/univiz/CollegeScoreCardData/CollegeScoreCardInterface.java
```diff
@@ -0,0 +1,16 @@
+import java.util.List;
+
+interface CollegeScoreCardInterface {
+  /*General Use Functions*/
+  public boolean validateCollegeName(String collegeName);
+  public List<UniversityData> getCollegesData(List<String> collegeNames);
+
+  /*Search Feature Functions*/
+  public List<UniversityData> getCollegeSuggestions(String currName);
```
biancamacias, line 9: is this a list of the colleges suggested for the autocomplete? or is this the list of colleges that the user selects? if this is not, then i suggest adding a ```public List<UniversityData> getSelectedColleges``` method.

mmurray22: these are the suggestions the autocomplete will provide the user; the getCollegesData function gets the list of colleges the user selects

capstone/src/main/java/com/google/univiz/CollegeScoreCardData/CollegeScoreCardInterface.java
```diff
@@ -0,0 +1,16 @@
...
+
+  /*Maps Feature Functions*/
+  public List<UniversityData> updateMapData(List<UniversityData> collegeDatum);
```
ihsan314, line 13: If this is just returning the data that will be used to create the map then maybe call it `getMapData`.

mmurray22: thanks! done!

capstone/src/main/java/com/google/univiz/CollegeScoreCardData/CollegeScoreCardInterface.java
biancamacias: nice ok. you dont have to do this but maybe making the method name ```getUserCollegesData``` or ```getSelectedCollegesData``` would make it clearer that this would be a list of the colleges that the user has picked.

capstone/src/main/java/com/google/univiz/CollegeScoreCardData/CollegeScoreCardInterfaceImpl.java
```diff
@@ -0,0 +1,65 @@
...
+import java.io.InputStreamReader;
+import java.io.Reader;
+import java.net.URL;
+import java.nio.charset.Charset;
+import org.json.JSONException;
+import org.json.JSONObject;
+
+public class CollegeScoreCardInterfaceImpl {
+  //TODO: THIS IS A PLACEHOLDER-REPLACE WITH REAL KEY//
+  private static final String apiKeyForCollegeScoreCard = "api_key=yeet:^)";
```
ihsan314: This is legendary.

capstone/src/main/java/com/google/univiz/CollegeScoreCardData/CollegeScoreCardInterfaceImpl.java
```diff
@@ -0,0 +1,65 @@
...
+     * -Each college has a UNIQUE name
+     * -Any query which returns multiple college
+     *  names will be marked as invalid
+     */
+    JSONObject collegeNameJson = getCollegeNameJson(collegeName); 
+    return (collegeNameJson.metadata.total == 1);
+  }
+
+  public List<UniversityData> getCollegesData(List<String> collegeNames) {
+    List<UniversityData> collegeData = new List<>();
```
ihsan314:
```suggestion
    List<UniversityData> collegeData = new ArrayList<>();
```
since List is an interface and cannot be instantiated. Also we'll see how things go, but it's very possible that college data may be too big for a `List`. But for now we can ignore that last part.

capstone/src/main/java/com/google/univiz/CollegeScoreCardData/CollegeScoreCardInterface.java
```diff
@@ -0,0 +1,17 @@
+import java.util.List;
```
ccondit, line 1: doesn't this interface need a package declaration?

capstone/src/main/java/com/google/univiz/CollegeScoreCardData/CollegeScoreCardInterface.java
```diff
@@ -0,0 +1,17 @@
...
+
+interface CollegeScoreCardInterface {
+  /*General Use Functions*/
+  public boolean validateCollegeName(String collegeName);
```
ccondit, line 5: You should be able to omit public on all interface methods. They [default to public visibility](http://docs.oracle.com/javase/specs/jls/se7/html/jls-6.html#jls-6.6.1).

capstone/src/main/java/com/google/univiz/CollegeScoreCardData/CollegeScoreCardInterface.java
```diff
@@ -0,0 +1,17 @@
+import java.util.List;
+
+interface CollegeScoreCardInterface {
```
ccondit, line 3: Is this the CollegeScoreCardInterface? Or is this a collection of interfaces that will have a CollegeScoreCard implementation? Considering refactoring this into multiple interfaces:

* CollegeAutocomplete
* MapData
* CollegeData

which in turn can have CollegeScoreCard implementation(s).

capstone/src/main/java/com/google/univiz/CollegeScoreCardData/CollegeScoreCardInterface.java
mmurray22: whoops looks like I left that off! sorry about that-done!

capstone/src/main/java/com/google/univiz/CollegeScoreCardData/CollegeScoreCardInterface.java
mmurray22: done!

capstone/src/main/java/com/google/univiz/CollegeScoreCardData/CollegeScoreCardInterface.java
mmurray22: I was hoping to keep all these functions in the same interface so that only one library will be calling the CollegeScoreCard API. Should I still refactor?

ccondit:
> NOTE: Will not pass Java Maven CI test until merged with the universityDataClass branch

Why is the CI test not even running?

capstone/src/main/java/com/google/univiz/CollegeScoreCardData/CollegeScoreCardInterface.java
```diff
@@ -0,0 +1,19 @@
+package com.google.univiz.collegescorecarddata;
```
ccondit, line 1: Note that the package name is correct (all lowercase), but the path of this interface is incorrect (`.../CollegeScoreCardData/...`).

Also, I suspect you'll not want your API interfaces in the `CollegeScoreCardData` package - since they shouldn't be inherently tied to CollegeScoreCard.

capstone/src/main/java/com/google/univiz/CollegeScoreCardData/CollegeScoreCardInterface.java
```diff
@@ -0,0 +1,19 @@
...
+
+import java.util.List;
+
+interface CollegeScoreCardInterface {
```
ccondit, line 5: In general, I don't think you need to suffix the name of interfaces with `Interface`. Users shouldn't really care if it's an interface or a class.

capstone/src/main/java/com/google/univiz/CollegeScoreCardData/CollegeScoreCardInterfaceImpl.java
mmurray22: haha

capstone/src/main/java/com/google/univiz/CollegeScoreCardData/CollegeScoreCardInterface.java
ccondit: Let's consider the autocomplete methods. Right now, this interface is tightly coupled to the entire surface of your API. What happens if you wanted to implement a better autocomplete later (that might have no knowledge of MapData or College Data)? If you had an CollegeAutocomplete interface agnostic of the rest of the signature, then it would be simple for you to wire up a new implementation.

capstone/src/main/java/com/google/univiz/CollegeScoreCardData/CollegeScoreCardInterface.java
```diff
@@ -0,0 +1,17 @@
...
+  /*General Use Functions*/
+  public boolean validateCollegeName(String collegeName);
+
+  public List<UniversityData> getCollegesData(List<String> collegeNames);
+
+  /*Search Feature Functions*/
```
ccondit, line 11: I'm not sure why the Java linter isn't working, but the formatting on these comments looks off (there should be spaces between the comment delimiters and the comment). Anyhow - these public methods should have [Javadoc](https://google.github.io/styleguide/javaguide.html#s7-javadoc). Can you add that?

capstone/src/main/java/com/google/univiz/CollegeScoreCardData/CollegeScoreCardInterface.java
ccondit, line 5: You'll probably want the interface(s) themselves to be public so you can access them outside of whatever package they end up in.

capstone/src/main/java/com/google/univiz/CollegeScoreCardData/CollegeScoreCardInterface.java
mmurray22: moved it! Thanks!

capstone/src/main/java/com/google/univiz/CollegeScoreCardData/CollegeScoreCardInterface.java
mmurray22: ah ok-updated!

mmurray22: done!

capstone/src/main/java/com/google/univiz/CollegeScoreCardData/CollegeScoreCardInterface.java
mmurray22: done! I think I did it right-let me know!

mmurray22:
How does the interface breakdown look?

capstone/src/main/java/com/google/univiz/collegescorecarddata/CollegeScoreCard.java
```diff
@@ -0,0 +1,19 @@
+package com.google.univiz.collegescorecarddata;
+
+import java.util.List;
+
+public interface CollegeScoreCard {
+  /** General Use Functions */
```
ccondit, line 6: what does this comment mean? what would non-general user function be?

capstone/src/main/java/com/google/univiz/collegescorecarddata/CollegeScoreCard.java
```diff
@@ -0,0 +1,19 @@
...
+  boolean validateCollegeName(String collegeName);
```
ccondit, line 7: What does this method do? How would an implementor know if a college name is valid?

capstone/src/main/java/com/google/univiz/collegescorecarddata/CollegeScoreCard.java
```diff
@@ -0,0 +1,19 @@
+package com.google.univiz.collegescorecarddata;
```
ccondit, line 1: Again - I'm not sure that you want this API interface in a `collegescorecarddata` package. If the backing implementation changes - we may not use collegescorecard at all...

capstone/src/main/java/com/google/univiz/collegescorecarddata/CollegeScoreCard.java
```diff
@@ -0,0 +1,19 @@
...
+import java.util.List;
+
+public interface CollegeScoreCard {
+  /** General Use Functions */
+  boolean validateCollegeName(String collegeName);
+
+  List<UniversityData> getCollegesData(List<String> collegeNames);
+
+  /** Search Feature Functions */
+  List<UniversityData> getCollegeSuggestions(String currName);
```
ccondit, line 12: should this be in the SearchData interface? actually looks like methods you refactored into other interfaces need to get removed from this interface...

capstone/src/main/java/com/google/univiz/collegescorecarddata/MapData.java
```diff
@@ -0,0 +1,8 @@
+package com.google.univiz.collegescorecarddata;
+
+import java.util.List;
+
+public interface MapData {
+  /** Maps Feature Functions */
+  List<UniversityData> getMapData(List<UniversityData> collegeDatum);
```
ccondit, line 7: what does it mean for a `List<UniversityData>` to produce a `List<UniversityData>`?

capstone/src/main/java/com/google/univiz/collegescorecarddata/SearchData.java
```diff
@@ -0,0 +1,8 @@
+package com.google.univiz.collegescorecarddata;
+
+import java.util.List;
+
+public interface CollegeScoreCard {
```
ccondit, line 5: This interface name is wrong (and actually won't build).

capstone/src/main/java/com/google/univiz/collegescorecarddata/SearchData.java
```diff
@@ -0,0 +1,8 @@
...
+  /** Search Feature Functions */
+  List<UniversityData> getCollegeSuggestions(String currName);
```
ccondit, line 7: Some thoughts here:
* what does the `curr` in `currName` mean?
* Based on the scope of `UniversityData` is shaping up - this is going to pull back a lot of data. Per this [API guidance](https://github.com/RTICWDT/open-data-maker/blob/master/API.md#limiting-returned-fields-with-fields) - we should be heavily restricting fields that come back from the server. I think there should be a light weight class for autocomplete that doesn't have all of the data that UniversityData has.

capstone/src/main/java/com/google/univiz/collegescorecarddata/MapData.java
mmurray22: made some comments on the new file!

capstone/src/main/java/com/google/univiz/collegescorecarddata/CollegeScoreCard.java
mmurray22: yes-fixed that!

capstone/src/main/java/com/google/univiz/collegescorecarddata/CollegeScoreCard.java
mmurray22: changed!

capstone/src/main/java/com/google/univiz/collegescorecarddata/CollegeScoreCard.java
mmurray22: updated comments!

capstone/src/main/java/com/google/univiz/collegescorecarddata/CollegeScoreCard.java
mmurray22: update comments!

capstone/src/main/java/com/google/univiz/collegescorecarddata/SearchData.java
mmurray22: yep-sorry for that oversight!

capstone/src/main/java/com/google/univiz/collegescorecarddata/SearchData.java
mmurray22: ah I see-I can just return a list of college names

capstone/src/main/java/com/google/univiz/api/MapData.java
```diff
@@ -0,0 +1,15 @@
...
+/**
+ * This interface is responsible for all data management/processing specific for the maps feature of
+ * the UniViz webapp.
+ */
+public interface MapData {
+  /**
+   * Takes a list of UniversityData objects and populates all fields relevant to the maps feature in
+   * each object; This function returns that same list withthe updated data fields
+   */
+  List<UniversityData> getMapData(List<UniversityData> collegeDatum);
```
kaveh096, line 14: @ccondit WDYT about returning immutable types? Worth it?

capstone/src/main/java/com/google/univiz/api/GeneralCollegeData.java
```diff
@@ -0,0 +1,21 @@
+package com.google.univiz.api;
```
kaveh096, line 1: This interface now belongs to a different package like `package com.google.univiz.scorecard;`

capstone/src/main/java/com/google/univiz/api/SearchData.java
```diff
@@ -0,0 +1,12 @@
+package com.google.univiz.api;
+
+import java.util.List;
+
+public interface SearchData {
+  /**
+   * This function take a potential college name fragment (provided by the user) and uses the API in
+   * use to determine whether this string fragment is present in any college names within the API
+   * database; If so, then a list of college name containing the provided string fragment
```
kaveh096, line 9:
```suggestion
   * database; If so, then returns a list of college name containing the provided string fragment.
```

capstone/src/main/java/com/google/univiz/api/SearchData.java
```diff
@@ -0,0 +1,12 @@
...
+   */
```
kaveh096, line 10: Also document if the returned data is sorted in any way.

capstone/src/main/java/com/google/univiz/api/VisData.java
```diff
@@ -0,0 +1,15 @@
+package com.google.univiz.api;
+
+import java.util.List;
+
+/**
+ * This interface is responsible for all data management/processing specific for the visualization
```
kaveh096, line 6: OOC, when you say 'visualization' are we talking about the charts? Cause mapping also could be considered a visualization. Maybe point out that it's the non-geographic visualizations. Or use other terms to describe it like chart or graph or tabular.

capstone/src/main/java/com/google/univiz/api/VisData.java
```diff
@@ -0,0 +1,15 @@
...
+ * feature of the UniViz webapp.
+ */
+public interface VisData {
+  /**
+   * Takes a list of UniversityData objects and populates all fields relevant to the visualization
+   * feature in each object; This function returns that same list with the updated data fields
+   */
+  List<UniversityData> getVisData(List<UniversityData> collegeDatum);
```
kaveh096, line 14: ditto: needs its on return data type.

PR #34, opened 2020-07-31
UniversityData.class implementation

biancamacias:
[this](https://github.com/googleinterns/univiz/pull/34/commits/9690279d17063adba5f2d9651bdabe737e902aa9) is a different approach to implementing the UniversityData.class without the builder pattern

### DIFF
--- a/capstone/src/main/java/com/google/univiz/UniversityData.java
+++ b/capstone/src/main/java/com/google/univiz/UniversityData.java
@@ -89,7 +89,22 @@ public final class UniversityData {
     }
 
     public UniversityData build() {
-      return new UniversityData(this);
+      UniversityData uni = new UniversityData();
+      uni.uniName = this.uniName;
+      uni.uniCity = this.uniCity;
+      uni.latitude = this.latitude;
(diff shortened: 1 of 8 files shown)


capstone/src/main/java/com/google/univiz/UniversityData.java
```diff
@@ -0,0 +1,105 @@
+package com.google.univiz;
+
+public final class UniversityData {
```
ccondit, line 3: Consider leveraging [AutoValue](https://github.com/google/auto/blob/master/value/userguide/index.md) for this class. This has a number of advantages and will prevent pitfalls (like not overriding hashcode, equals, etc). See also Effective Java chapter 3.

I'm adding a dependency on AutoValue in https://github.com/googleinterns/univiz/pull/33 - so feel free to rebase off of that branch or wait until I merge...

ccondit: Did this get addressed? I think it will improve this class.

biancamacias:
got the changes made to ```master```, now im getting these errors

image

Issue #26, opened 2020-07-30
Adds CollegeScoreCard API Library


mmurray22:
This issue addressed the need for a CollegeScoreCard API Library which will provide easy-to-use functions for the Search, Maps, and DataVis servlets to use when enacting their respective functionalities. Step I will take to address this issue will be the following:
> Create a Java class focused on being the CollegeScoreCard API Library

> Create a series of functions which coordinate well with the other servlets

> Make calls to the API within these functions So no other set let had to directly interact with the API

Issue #23, opened 2020-07-30
Graphing College Tuition Data For UniViz.


ihsan314:
One of the essential features of UniViz is graphing tuition data coming from the CollegeScoreCard API. 

In this [doc](http://go/univiz-data-viz) I outlined the way I intend to set out on adding a graph for financial data to UniViz. There I stated that I will add a bar graph comparing the user's budget with gross/net cost of tuition.

Here are the steps I will take to do that:

- [x]  Deserialize the JSON object inside the `graphs.js` file
- [x]  Add this data to a [DataTable](https://developers-dot-devsite-v2-prod.appspot.com/chart/interactive/docs/reference#DataTable)
- [x]  Set up a double bar graph with tuition as the y-axis and the user's colleges as the x-axis, and the gross/net tuition as the two sets of bars per college, using the DataTable
- [x]  Add a `<div>` element that will be containing the chart

PR #35, opened 2020-07-31
Simple bar graph

ihsan314:
This addresses the latter two tasks of #23, and sets up for the first two tasks. Depending on what comes out of #26, the first two tasks of #23 could vary significantly. As a result the first two tasks will be addressed in another PR.


capstone/src/main/webapp/index.html
```diff
@@ -11,5 +13,6 @@ <h2>Welcome to UniViz!</h2>
       <input type="text" id="search" name="search"><br>
       <input type="submit" value="See your colleges!">
     </form>
+    <div id="data"></div>
```
biancamacias, line 16: just note that the data ```<div>``` will be on the dashboard.html page and not the index.html page

ihsan314: Ah ok, when that gets merged into master I'll move the `<div>` there.

ihsan314:
I thought it would be simpler to have individual javascript files for each thing, but if we want to have one giant javascript file with all the javascript functionality we need then that works too.

ihsan314:
I'm thinking of tests to add, but I didn't really write this in a test-friendly manner. I see one aspect that can be modularized for testing, which is the deserialization of the JSON object and populating the DataTable. (This aspect is probably very similar for my [other JavaScript pull request](https://github.com/googleinterns/univiz/pull/39) except there are two DataTables instead of one.) I haven't added this functionality yet, so I will add that. Then I'll refactor and test this. If there's anything wrong with my plan let me know.

capstone/src/main/webapp/js/graphs.js
```diff
@@ -0,0 +1,30 @@
...
+ * The callback function that will generate the double
+ * bar graph. Uses Google Charts API and JSON object
+ * consisting of college tuition data.
+ */
+function graphGrossNetTuition() {
+  const data = new google.visualization.DataTable();
+  fetch('/financial-data').
+      then((response) => response.json()).
+      then((tuitionInfo) => {
+        // iterate over tuitionInfo JSON
```
kaveh096, line 14: This comment reads like a TODO. If that is the case then start your comment with `// TODO(github_issue_id): Iterate over tuitionInfo JSON...`

capstone/src/test/javascript/graphsTest.js
```diff
@@ -0,0 +1,17 @@
+describe('Data Table Construction', function() {
+  it('Will add correct columns/rows to DataTable from JSON', function() {
+    // TODO(ihsan314): Change this when format of JSON object is well-defined
+    const script = document.createElement('script');
+    script.src = 'https://www.gstatic.com/charts/loader.js';
+    script.defer = true;
```
ccondit, line 6: Are these lines necessary? Looks like the test passes without them...

capstone/src/test/javascript/graphsTest.js
```diff
@@ -0,0 +1,17 @@
+describe('Data Table Construction', function() {
+  it('Will add correct columns/rows to DataTable from JSON', function() {
+    // TODO(ihsan314): Change this when format of JSON object is well-defined
```
ccondit, line 3: What does `this` mean in the `TODO`?

capstone/src/test/javascript/graphsTest.js
```diff
@@ -0,0 +1,17 @@
...
+    const script = document.createElement('script');
+    script.src = 'https://www.gstatic.com/charts/loader.js';
+    script.defer = true;
+
+    window.initMap = function() {
+      const mockTuitionInfo = {};
+      const expectedData = new google.visualization.DataTable();
+      const actualData = new google.visualization.DataTable();
+      populateDataTable(actualData, mockTuitionInfo);
+      expect(actualData).toBe(expectedData);
```
ccondit, line 13: actually - i think this `expect` assertion isn't even being invoked since it's in a function. if you put jiberish in here does the test still pass?

capstone/src/main/webapp/js/graphs.js
```diff
@@ -0,0 +1,61 @@
...
+   *
+   * @return {google.visualization.DataTable} data
+   * the DataTable corresponding to the tuition data
+   */
+  function deserializeTuitionData() {
+    const data = new google.visualization.DataTable();
+    fetch('/tuition-data')
+        .then((response) => response.json())
+        .then((tuitionInfo) => populateDataTable(data, tuitionInfo));
+    return data;
```
ccondit, line 41: I'm curious if these will even test since they're defined inside of your callback. Do we need to move them to the top level of this file in order to get tests working?

capstone/src/main/webapp/js/graphs.js
```diff
@@ -0,0 +1,61 @@
+const script = document.createElement('script');
+script.src = 'https://www.gstatic.com/charts/loader.js';
+script.defer = true;
+
+window.initMap = function() {
```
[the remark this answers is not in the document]
ihsan314: Oh ok, will change it.

capstone/src/test/javascript/graphsTest.js
ihsan314: I think tests weren't passing when I didn't have those lines. I'll check again and confirm.

capstone/src/test/javascript/graphsTest.js
ihsan314: I'll check this too, maybe I placed the `expect` assertion incorrectly.

capstone/src/main/webapp/js/graphs.js
ihsan314: Maybe, this is yet another thing I will check.

capstone/src/test/javascript/graphsTest.js
ihsan314: The `TODO` makes more sense with the definition of the `JSON` object and the `DataTables`. I'll move it there.

capstone/src/test/javascript/graphsTest.js
ihsan314: Yea I confirmed tests are ok without them.

capstone/src/test/javascript/graphsTest.js
ihsan314: It's correct that the `expect` wasn't getting invoked, I'll move it out of the callback.

capstone/src/main/webapp/js/graphs.js
```diff
@@ -0,0 +1,61 @@
+const script = document.createElement('script');
+script.src = 'https://www.gstatic.com/charts/loader.js';
+script.defer = true;
+
+window.initTuitionData = function() {
```
fabielzuniga, line 5: Google favors arrow function:
() => {
  ...
}

capstone/src/main/webapp/js/graphs.js
```diff
@@ -0,0 +1,61 @@
...
+   *
+   * @return {google.visualization.DataTable} data
+   * the DataTable corresponding to the tuition data
+   */
+  function deserializeTuitionData() {
+    const data = new google.visualization.DataTable();
+    fetch('/tuition-data')
+        .then((response) => response.json())
+        .then((tuitionInfo) => populateDataTable(data, tuitionInfo));
+    return data;
```
fabielzuniga, line 41: Why do we return the data? I think fetch('/tuition-data') will execute asynchronously so when we return data the table might have not been initialized yet.

capstone/src/test/javascript/graphsTest.js
```diff
@@ -0,0 +1,15 @@
+describe('Data Table Construction', function() {
```
fabielzuniga: Same comment about using arrow function. Same comment for other instances below.

capstone/src/test/javascript/graphsTest.js
```diff
@@ -0,0 +1,15 @@
...
+  it('Will add correct columns/rows to DataTable from JSON', function() {
+    const expectedData;
```
fabielzuniga: This is marked as const and then its value is set later. 
Same comment for the other fields.

capstone/src/main/webapp/js/graphs.js
```diff
@@ -0,0 +1,61 @@
...
+    const chart =
+      new google.visualization.ColumnChart(document.getElementById('data'));
+    chart.draw(data, options);
+  }
+
+  /**
+   * Helper function to fetch elements from JSON servlet.
+   *
+   * @return {google.visualization.DataTable} data
+   * the DataTable corresponding to the tuition data
```
fabielzuniga, line 34: Check the accepted line wrapping for JavaScript doc tags:
https://google.github.io/styleguide/jsguide.html#jsdoc-line-wrapping

Same comment for @param

capstone/src/test/javascript/graphsSpec.js
```diff
@@ -0,0 +1,26 @@
...
+    let actualData = {};
+    const mockTuitionInfo = {};
+    window.initTuitionData = () => {
+      // TODO(ihsan314): Initialize this when format of
+      // JSON object is well-defined
+      expectedData = new google.visualization.DataTable();
+      actualData = new google.visualization.DataTable();
+      console.log(expectedData);
+      console.log(actualData);
+      populateDataTable(actualData, mockTuitionInfo);
```
ccondit, line 22: I don't think this block of code is ever invoked. 

And I think this should test `deserializeTuitionData` instead of populateDataTable. That way we can mock our response and make sure it converts correctly...

ihsan314: Since I'm mocking it anyway I don't need the `initTuitionData` callback so I'll remove it. Additionally I'll change the test to test `deserializeTuitionData`.

ihsan314:
Ok I think all should be ready, although the test is timing out for some reason. I don't understand why though, since I spied on the fetch.

capstone/src/test/javascript/graphsSpec.js
```diff
@@ -0,0 +1,56 @@
...
+   * Dummy function for adding columns to mock DataTable class.
+   *
+   * @param {String} type the datatype of the column.
+   * @param {String} name the name of the column.
+   */
+  addColumn(type, name) {}
+};
+
+
+describe('Data Table Construction', () => {
```
fabielzuniga, line 26: I just notice this is part of a file named graphSpec. Should we add it to a test file? For example graph_test.js

ihsan314: The naming convention for the test files with the Jasmine framework is to use `Spec` at the end, otherwise I would have named it with `test` or something similar.

fabielzuniga: I see, sounds good.

PR #37, opened 2020-07-31
ScorecardData.class implementation

biancamacias:
this PR uses ```@AutoValue``` for the class implementation

this class will be used to organize the data from the CollegeScorecard API, where each university will be of type ```ScorecardData```. when the servlets interact with the selected colleges, they'll be using ```List<ScorecardData>```. This class also uses ```@SerializedName``` in order to parse the JSON data from the API correctly.

This PR also contains the ```ScorecardTest.java``` file, where all the unit tests for this class will live. The tests also check if null values will parse correctly, as the example ```scorecard.json``` has a ```null``` ```urbanizationDegree()```.


capstone/src/main/java/com/google/univiz/UniversityData.java
```diff
@@ -0,0 +1,67 @@
...
+        flagMainCampus,
+        urbanizationDegree,
+        admissionRate,
+        avgSAT,
+        numOfUndergrads,
+        avgCost,
+        studentsInDebt,
+        carnegieSizeDegree,
+        numOfMen,
+        numOfWomen);
```
ihsan314, line 37: Calling the UniversityData constructor doesn't work for abstract classes. Even if it did, default constructors take no arguments.

capstone/src/main/java/com/google/univiz/UniversityData.java
```diff
@@ -0,0 +1,67 @@
...
+
+  public static UniversityData create(
+      String name,
+      String city,
+      float latitude,
+      float longitude,
+      int flagMainCampus,
+      int urbanizationDegree,
+      float admissionRate,
+      float avgSAT,
```
ccondit, line 16: per style, please use [lower camel case](https://google.github.io/styleguide/javaguide.html#s5.3-camel-case) (and in the accessor method, as well)

capstone/src/main/java/com/google/univiz/UniversityData.java
```diff
@@ -0,0 +1,67 @@
+package com.google.univiz;
+
+import com.google.auto.value.AutoValue;
+
+@AutoValue
+public abstract class UniversityData {
```
ccondit, line 6: nit - add [javadoc for public classes](https://google.github.io/styleguide/javaguide.html#s5.3-camel-case)

capstone/src/main/java/com/google/univiz/UniversityData.java
```diff
@@ -0,0 +1,67 @@
...
+
+  public static UniversityData create(
+      String name,
+      String city,
```
ccondit, line 10: Does it make sense to include the entire address?

capstone/src/main/java/com/google/univiz/UniversityData.java
```diff
@@ -0,0 +1,67 @@
...
+      float latitude,
+      float longitude,
```
ccondit, line 12: nit - it might make sense to introduce another class to represent location. maybe it contains an address and a lat lon? and then that class is an argument to this `create` method.

capstone/src/main/java/com/google/univiz/UniversityData.java
```diff
@@ -0,0 +1,67 @@
...
+      int flagMainCampus,
```
ccondit, line 13: what does `flag` mean in this name? can this be better represented as a boolean? or a enum (if there are other types of campuses)?

capstone/src/main/java/com/google/univiz/UniversityData.java
```diff
@@ -0,0 +1,67 @@
...
+      int urbanizationDegree,
```
ccondit, line 14: what does urbanizationDegree mean? would definitely be worth adding javadoc about this...

capstone/src/main/java/com/google/univiz/UniversityData.java
```diff
@@ -0,0 +1,67 @@
...
+      float admissionRate,
+      float avgSAT,
+      int numOfUndergrads,
+      int avgCost,
+      int studentsInDebt,
+      int carnegieSizeDegree,
```
ccondit, line 20: carnegieSizeDegree doesn't mean much to the causal reader. also would benefit from javadoc.

capstone/src/main/java/com/google/univiz/UniversityData.java
```diff
@@ -0,0 +1,67 @@
...
+      float numOfMen,
+      float numOfWomen) {
```
ccondit, line 22: are there fractional numbers of men and women?

capstone/src/main/java/com/google/univiz/UniversityData.java
biancamacias: its not meant to be the entire address, just the name of the city

capstone/src/main/java/com/google/univiz/UniversityData.java
biancamacias: > are there fractional numbers of men and women?

ill bring up at standup why i chose to use these types for the variables

capstone/src/main/java/com/google/univiz/UniversityData.java
```diff
@@ -0,0 +1,90 @@
+package com.google.univiz;
```
kaveh096, line 1: Per the standup discussion, here is what we agreed this data structure should do:

 - It should capture the scorecard return data in its pure form with no extra interpretation.
 - It should have all the fields that our webapp would need for the mvp features and no extra fields.
 - The field names should match the scorecard field name so gson can deserialize correctly.

kaveh096, line 1: Move to a dedicated package like `package com.google.univiz.scorecard;`

capstone/src/main/java/com/google/univiz/UniversityData.java
```diff
@@ -0,0 +1,90 @@
...
+
+import com.google.auto.value.AutoValue;
+
+/** UniversityData is a class representing a single university. */
+@AutoValue
+public abstract class UniversityData {
```
kaveh096, line 7: Rename class to `ScorecardData`.

capstone/src/main/java/com/google/univiz/UniversityData.java
```diff
@@ -0,0 +1,90 @@
...
+   * @param numOfUndergrads The number of undergraduate students that attend the university.
+   * @param avgCost The average cost of attendance for students at the university.
+   * @param studentsInDebt The number of students in cumulative loan debt after graduation.
+   * @param carnegieSizeDegree An integer from range(-2-18) that tells degree of Carnegie size, from
+   *     not applicable to exclusively graduate/professional.
+   * @param numOfMen A float that represents the number of men enrolled at the undergraduate level.
+   * @param numOfWomen A float that represents the number of women enrolled at the undergradute
+   *     level.
+   */
+  public static UniversityData create(
```
kaveh096, line 31: I imagine we won't ever need this factory method. Because we always want to build this class based on the scorecard api. So gson would be responsible for building this class and not this factory method. Here is [an example](https://sites.google.com/site/gson/gson-user-guide#TOC-Object-Examples).

@ccondit is `AutoValue` going to work with `gson` correctly? I think it needs a class with argument-less constructor.

ccondit: `AutoValue` should work with `gson`. I set this up in #33 for reading json config data from json into an autovalue.

capstone/src/main/java/com/google/univiz/UniversityData.java
ccondit: +1 - can you also add a unit test that verifies that json from the service can be parsed by your new autovalue class?

The sample json input can live in `src/test/resources/com/google/univiz/scorecard`.

capstone/src/main/java/com/google/univiz/UniversityData.java
```diff
@@ -0,0 +1,90 @@
+package com.google.univiz;
+
+import com.google.auto.value.AutoValue;
+
+/** UniversityData is a class representing a single university. */
+@AutoValue
+public abstract class UniversityData {
+
+  /**
+   * Creates a new UniversityData
```
mmurray22, line 10: Overall looking good :) there are a lot of comments already here, so I'll rereview when those comments are resolved and see if I have anything else to add

capstone/src/main/java/com/google/univiz/UniversityData.java
biancamacias: ive added a unit test in #48.

capstone/src/main/java/com/google/univiz/scorecard/ScorecardData.java
```diff
@@ -0,0 +1,108 @@
...
+
+/**
+ * ScorecardData is a class representing a single university/college/institution from the parsed
+ * JSON data directly from the CollegeScorecard API.
+ */
+@AutoValue
+public abstract class ScorecardData {
+
+  /**
+   * Creates a new ScorecardData
```
ccondit, line 14: It would be worth referencing where these param descriptions are coming from. That way if someone needs to add a field in the future it will be easy to find...

capstone/src/main/java/com/google/univiz/scorecard/ScorecardData.java
```diff
@@ -0,0 +1,108 @@
...
+
+import com.google.auto.value.AutoValue;
+import com.google.gson.annotations.SerializedName;
+
+/**
+ * ScorecardData is a class representing a single university/college/institution from the parsed
+ * JSON data directly from the CollegeScorecard API.
+ */
+@AutoValue
+public abstract class ScorecardData {
```
ccondit, line 11: some visibility nits - since this class will only be used in this package by another class that will parse the json, it would be a good idea to make every method in this class (and the class, itself) package private. This will prevent it from leaking out past this package.

TLDR: s/public//g

capstone/src/main/java/com/google/univiz/UniversityData.java
ccondit: Sounds good - FWIW going forward it will be nice to have tests in the same PR as the class under test. That way we can verify that the test WAI and can request new test cases while reviewing the code...

ccondit: (I'm also fine with you pulling the test into this PR)

biancamacias: ohhhh okay. that makes sense. i will make sure to put tests in the same PR in the future instead of pulling separate ones. (ill merge them though!)

capstone/src/main/java/com/google/univiz/scorecard/ScorecardData.java
```diff
@@ -8,10 +8,12 @@
...
  */
 @AutoValue
-public abstract class ScorecardData {
+private abstract class ScorecardData {
 
   /**
-   * Creates a new ScorecardData
+   * Creates a new ScorecardData. The parameters come directly from the CollegeScorecard API Data
+   * Dictionary, which outlines the type and description of all available data elements. The
+   * parameters used here are a select few data elements the API provides.
```
ccondit, line 16: Can you link to 
https://collegescorecard.ed.gov/assets/CollegeScorecardDataDictionary.xlsx
in the javadoc? That way future modifiers will know where things came from.

capstone/src/main/java/com/google/univiz/scorecard/ScorecardData.java
biancamacias: its saying im not allowed to make the class private? not sure whats happening.

capstone/src/main/java/com/google/univiz/scorecard/ScorecardData.java
```diff
@@ -8,10 +8,12 @@
  * JSON data directly from the CollegeScorecard API.
  */
 @AutoValue
-public abstract class ScorecardData {
+private abstract class ScorecardData {
```
ccondit: Everything (including this class) should be package private. See [this access control documentation](https://docs.oracle.com/javase/tutorial/java/javaOO/accesscontrol.html). If we make things private then they won't be usable by other classes the scorecard package.

capstone/src/main/java/com/google/univiz/scorecard/ScorecardData.java
biancamacias: for the methods, i can remove public from it, but [i cant make them private](https://github.com/google/auto/blob/master/value/userguide/howto.md#-keep-my-accessor-methods-private). for the class, i think i can only remove public from it as well.

capstone/src/main/java/com/google/univiz/scorecard/ScorecardData.java
```diff
@@ -8,10 +8,12 @@
  * JSON data directly from the CollegeScorecard API.
  */
 @AutoValue
```
ccondit, line 10: To get this to work with gson you'll also need to annotate this class with `@GenerateTypeAdapter`.

capstone/src/main/java/com/google/univiz/scorecard/ScorecardData.java
```diff
@@ -32,7 +34,7 @@
    * @param numOfWomen A float that represents the number of women enrolled at the undergradute
    *     level.
    */
-  public static ScorecardData create(
+  private static ScorecardData create(
```
ccondit, line 37: Also - should have caught this earlier, do we actually need the `create` method yet? I don't think GSON is going to use this. We will want to create instance of this class by hand when we test translating it into our domain object - but I think will be more readable with [builders](https://github.com/google/auto/blob/master/value/userguide/builders.md) - and we can add them when we get there.

In summary let's:
* Remove the create method entirely
* Migrate the @param javadocs to the individual accessor methods 
* Add that note about the URL where the descriptions came from in the Javadoc for the class

ccondit: PS - this is why having the tests in the same PR is sometimes helpful. We could verify that removing the create method is ok (I'm pretty sure it is 😃 ).

capstone/src/main/java/com/google/univiz/scorecard/ScorecardData.java
ccondit: See my recent comments. You'll want to use package private for things here (which means no access modifier)

capstone/src/main/java/com/google/univiz/scorecard/ScorecardData.java
biancamacias: i dont think autovalue doesnt support this. the documentation says autovalue methods at least needs to be visible by the package, so i can remove ```public``` but i cant add private, and i think it applies to the class as well because when i add ```private``` here, it doesnt build.

ccondit: Right - I'm suggesting that this class should be `package private` (without a modifier)

capstone/src/main/java/com/google/univiz/scorecard/ScorecardData.java
```diff
@@ -0,0 +1,79 @@
...
+@GenerateTypeAdapter
+@AutoValue
+abstract class ScorecardData {
+
+  /**
+   * Creates a new ScorecardData. The parameters come directly from the CollegeScorecard API Data
+   * Dictionary (https://collegescorecard.ed.gov/assets/CollegeScorecardDataDictionary.xlsx), which
+   * outlines the type and description of all available data elements. The parameters used here are
+   * a select few data elements the API provides.
+   */
```
ccondit, line 20: nit - please merge this comment with the class Javadoc.

capstone/src/main/java/com/google/univiz/scorecard/ScorecardData.java
biancamacias: the test has been moved into this PR :)

capstone/src/test/resources/com/google/univiz/scorecard/ScorecardTest.java
```diff
@@ -0,0 +1,37 @@
+package com.google.univiz.scorecard;
```
ccondit: This class should be in `src/test/java/com/google/univiz/scorecard`. As it is, your tests aren't getting invoked by Maven. Can you move it and then get the test to pass?

capstone/src/main/java/com/google/univiz/scorecard/ScorecardData.java
```diff
@@ -0,0 +1,76 @@
...
+ * CollegeScorecard API Data Dictionary
+ * (https://collegescorecard.ed.gov/assets/CollegeScorecardDataDictionary.xlsx), which outlines the
+ * type and description of all available data elements. The parameters used here are a select few
+ * data elements the API provides.
+ */
+@GenerateTypeAdapter
+@AutoValue
+abstract class ScorecardData {
+
+  /** @return The human-readable name for the school/college/university. */
```
kaveh096, line 19: The first line of a method javadoc should be a summary of its behavior. If describing the return object is the summary explanation then don't skip the summary in favor of the @return section. Just say "Returns the human-readable..." and then skip the @return. You only need the @return line if you want to further clarify details about the returned object. Like for example whether the method returns null or whether it returns an empty string or not.

Take a look at [javadoc.pdf page 5](http://dr/corp/drive/folders/14pg9M6eoAC_yM1XjC2IuocxX80R0lMXu).

capstone/src/main/java/com/google/univiz/scorecard/ScorecardData.java
```diff
@@ -0,0 +1,76 @@
...
+  /** @return The human-readable name for the city the university is located in. */
+  @SerializedName("school.city")
+  abstract String city();
+
+  /** @return An integer that flags whether the university is the main campus or not, 1 or 0. */
+  @SerializedName("school.main_campus")
+  abstract int flagMainCampus();
+
+  /**
+   * @return An integer from range(1-8) that tells the degree of urbanization for that campus, from
```
kaveh096, line 32: Does this range mean 1-8 inclusive or exclusive? Consider using [this interval notation](https://en.wikipedia.org/wiki/Interval_(mathematics)#Notations_for_intervals) to properly document the ranges. For example if the range includes 1 but excludes 8, you can say `range [1, 8)`.

capstone/src/main/java/com/google/univiz/scorecard/ScorecardData.java
```diff
@@ -0,0 +1,76 @@
...
+   *     large city to rural.
+   */
+  @SerializedName("school.degree_urbanization")
+  abstract int urbanizationDegree();
+
+  /** @return A float representing the school's location measured in latitude. */
```
kaveh096, line 38: Wording suggestion:
```suggestion
  /** Returns a float representing the school's latitude location measured in degrees. */
```

capstone/src/main/java/com/google/univiz/scorecard/ScorecardData.java
```diff
@@ -0,0 +1,76 @@
...
+  @SerializedName("location.lat")
+  abstract float latitude();
+
+  /** @return A float representing the school's location measured in longitude. */
```
kaveh096, line 42:
```suggestion
  /** Returns a float representing the school's longitude location measured in degrees. */
```

capstone/src/main/java/com/google/univiz/scorecard/ScorecardData.java
```diff
@@ -0,0 +1,76 @@
...
+  @SerializedName("location.lon")
+  abstract float longitude();
+
+  /**
+   * @return An integer from range(-2-18) that tells degree of Carnegie size, from not applicable to
```
kaveh096, line 47: ditto range notation

capstone/src/main/java/com/google/univiz/scorecard/ScorecardData.java
```diff
@@ -0,0 +1,76 @@
...
+
+  /** @return The number of undergraduate students that are enrolled the university. */
+  @SerializedName("2018.student.size")
+  abstract int numOfUndergrads();
+
+  /** @return The average cost of attendance for students per academic year at the university. */
+  @SerializedName("2018.cost.attendance.academic_year")
+  abstract int avgCost();
+
+  /** @return A float that represents the number of men enrolled at the undergraduate level. */
```
kaveh096, line 69:
```suggestion
  /** Returns a float that represents the ratio of men enrolled at the undergraduate level. */
```

capstone/src/main/java/com/google/univiz/scorecard/ScorecardData.java
```diff
@@ -0,0 +1,76 @@
...
+  @SerializedName("2018.student.demographics.men")
+  abstract float numOfMen();
+
+  /** @return A float that represents the number of women enrolled at the undergradute level. */
```
kaveh096, line 73:
```suggestion
  /** Returns a float that represents the ratio of women enrolled at the undergraduate level. */
```

capstone/src/test/java/com/google/univiz/scorecard/ScorecardTest.java
```diff
@@ -0,0 +1,29 @@
...
+@RunWith(JUnit4.class)
+public final class ScorecardTest {
+
+  private final ScorecardData scorecardData;
+
+  @Test
+  public void testNameDeserializes() {
+    Gson gson = new GsonBuilder().registerTypeAdapterFactory(GenerateTypeAdapter.FACTORY).create();
+    InputStreamReader scorecardReader =
+        new InputStreamReader(Resources.getResource("/resources/com/google/univiz/scorecard/scorecard.json").openStream()) {
```
ccondit, line 25:
```suggestion
InputStreamReader scorecardReader =
        new InputStreamReader(Resources.getResource(ScorecardTest.class, "scorecard.json").openStream());
```

this way it the path stays relative to the class.

capstone/src/test/java/com/google/univiz/scorecard/ScorecardTest.java
```diff
@@ -0,0 +1,29 @@
...
+import java.io.InputStreamReader;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public final class ScorecardTest {
+
+  private final ScorecardData scorecardData;
```
ccondit, line 19: just move this inline into the test...

capstone/src/test/java/com/google/univiz/scorecard/ScorecardTest.java
```diff
@@ -0,0 +1,29 @@
...
+
+  @Test
+  public void testNameDeserializes() {
+    Gson gson = new GsonBuilder().registerTypeAdapterFactory(GenerateTypeAdapter.FACTORY).create();
+    InputStreamReader scorecardReader =
+        new InputStreamReader(Resources.getResource("/resources/com/google/univiz/scorecard/scorecard.json").openStream()) {
+      return gson.fromJson(scorecardReader, ScorecardData.class);
+    assertThat(scorecardData.name().isEqualTo("New York University"));
```
ccondit, line 27:
```suggestion
    assertThat(scorecardData.name()).isEqualTo("New York University");
```

capstone/src/test/resources/com/google/univiz/scorecard/scorecard.json
```diff
@@ -0,0 +1,25 @@
+{
```
ccondit, line 1: Note that you'll probably need to introduce another Autovalue class to handle this full response:
* `ScorecardData` corresponds to a single `result`.
* The new Autovalue would look something like:
```
@AutoValue
SearchResponse {
  abstract List<ScorecardData> results;
}
```
* This can happen in a follow up PR

biancamacias:
im getting this error with ```mvn test``` currently:

image

capstone/src/test/resources/com/google/univiz/scorecard/scorecard.json
biancamacias: is this so that it doesnt read the ```metadata``` array?

capstone/src/test/java/com/google/univiz/scorecard/ScorecardTest.java
```diff
@@ -0,0 +1,30 @@
...
+import java.io.InputStreamReader;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public final class ScorecardTest {
+
+  @Test
+  public void testJsonDeserializes() {
```
ccondit, line 18:
```suggestion
  public void testJsonDeserializes() throws Exception {
```

ccondit:
> im getting this error with `mvn test` currently:
> 
> image

See: https://github.com/googleinterns/univiz/pull/37#discussion_r466567672

capstone/src/main/java/com/google/univiz/scorecard/ScorecardData.java
```diff
@@ -0,0 +1,91 @@
...
+import com.google.auto.value.AutoValue;
+import com.google.gson.annotations.SerializedName;
+import com.ryanharter.auto.value.gson.GenerateTypeAdapter;
+import javax.annotation.Nullable;
+
+/**
+ * ScorecardData is a class representing a single university/college/institution from the parsed
+ * JSON data directly from the CollegeScorecard API. The parameters come directly from the
+ * CollegeScorecard API Data Dictionary
+ * (https://collegescorecard.ed.gov/assets/CollegeScorecardDataDictionary.xlsx), which outlines the
```
ccondit, line 12: Just a heads up that there are other Javadoc constructs you can use for links - namely `<a href=""></a>`.

Just for future reference, no need to change anything now.

capstone/src/main/java/com/google/univiz/scorecard/ScorecardData.java
```diff
@@ -0,0 +1,91 @@
...
+  @SerializedName("school.city")
+  @Nullable
+  abstract String city();
+
+  /** Returns an integer that flags whether the institution is the main campus or not, 1 or 0. */
+  @SerializedName("school.main_campus")
+  abstract int flagMainCampus();
+
+  /**
+   * Returns an integer from range [1-8] that tells the degree of urbanization for institution's
```
kaveh096, line 41:
```suggestion
   * Returns an integer from range [1, 8] that tells the degree of urbanization for institution's
```

capstone/src/main/java/com/google/univiz/scorecard/ScorecardData.java
```diff
@@ -0,0 +1,91 @@
...
+  /** Returns a double representing the institution's latitude location measured in degrees. */
+  @SerializedName("location.lat")
+  abstract double latitude();
+
+  /** Returns a double representing the institution's longitude location measured in degrees. */
+  @SerializedName("location.lon")
+  abstract double longitude();
+
+  /**
+   * Returns an integer from range [-2-18] that tells degree of Carnegie size, from not applicable
```
kaveh096, line 56:
```suggestion
   * Returns an integer from range [-2, 18] that tells degree of Carnegie size, from not applicable
```

PR #43, opened 2020-08-03
Adds interface files

mmurray22:
This PR includes 3 files:
-MapData.java
-SearchData.java
-VisData.java
There is no content in these files right now as I would like this PR to focus on whether or not the 1) package name and 2) the names of the interface are explanatory enough. Eventually, these interfaces will contain functions that each feature's servlet will call when obtaining their data. For example, the MapData interface will contain a function which will return all API data that is specific to the MapData functionality.

Let me know if this PR needs more/or anything else!


mmurray22:
@biancamacias My understanding is I will set up the interfaces and then you can take over the interface functionality.

capstone/src/main/java/com/google/univiz/api/MapData.java
```diff
@@ -0,0 +1 @@
+package com.google.univiz.api;
```
kaveh096, line 1: i'm actually not sure if it's ok to have an empty `.java` file. @ccondit do you know? Other than that i'm ok with the package organization. Note that since you've named this file `MapData.java` technically you have already picked the interface name. So you might as well add it here:

```java
/** */
public interface MapData {}
```

mmurray22: a noted! will do!

ccondit: i think it would be fine, in theory. but i think adding the interface is fine in this PR.

mmurray22:
does the preliminary JavaDoc on the objects look sufficient?

PR #44, opened 2020-08-03
Adds basic content to the MapData Interface

mmurray22:
This PR creates a basic MapData Interface. I have yet to create the return object's class, but I wanted feedback on the name/the usefulness of the object before proceeding to go further with it. 

Here is a doc with further information about the [data flow framework](https://docs.google.com/document/d/1G2QrgDMhVCPX6cbNOQNKvexDImB5Duw1dr50cDGx70U/edit)


capstone/src/main/java/com/google/univiz/api/MapData.java
```diff
@@ -1 +1,13 @@
...
+/*
+ * The MapData interface is responsible for all data management/processing specific for the maps 
+ * feature of the UniViz webapp.
+ */
+public interface MapData {
+  /*
+   * Takes a list of college names and returns a CollegeMapsData object to the caller. This object 
+   * contains all data relevant to the Maps feature.
+   */
+  CollegeMapsData getMapData(List<String> collegeNames);
```
ihsan314, line 12: Why `CollegeMapsData` instead of `MapsData`?

capstone/src/main/java/com/google/univiz/api/SearchData.java
```diff
@@ -1 +1,3 @@
 package com.google.univiz.api;
+
+import java.util.List;
```
ihsan314: Possibly add the java doc for the SearchData interface or otherwise describe it in the PR.

mmurray22: took these changes away from searchdata :)

capstone/src/main/java/com/google/univiz/api/MapData.java
mmurray22: MapsData is the name of the interface so I didn't want to use the same name for the possible object name...

ihsan314: I think the general practice is to use the interface for the return type if possible to achieve maximal abstraction.

mmurray22: Hi Ihsan, I didn't quite understand what you meant by using the interface as the return type? Can you expand on that? Thank you!

mmurray22:
I will make a doc responding to these concerns and share it with the group and then we can go from there.

capstone/src/main/java/com/google/univiz/api/MapData.java
```diff
@@ -1 +1,13 @@
 package com.google.univiz.api;
+
+/*
+ * The MapData interface is responsible for all data management/processing specific for the maps
+ * feature of the UniViz webapp.
+ */
+public interface MapData {
+  /*
+   * Takes a list of college names and returns a CollegeMapsData object to the caller. This object
```
kaveh096, line 9: Should we use a more stable identifier for colleges? Something like a unique id (could be scorecard uid) instead of name.

capstone/src/main/java/com/google/univiz/api/MapData.java
```diff
@@ -1 +1,13 @@
...
+   * contains all data relevant to the Maps feature.
+   */
+  CollegeMapsData getMapData(List<String> collegeNames);
```
kaveh096, line 12: I expected to see the `CollegeMapsData` class definition in this PR too. If you're not sure about its fields, start by adding an empty class and fill it out in a followup PR.

kaveh096, line 12: Should the return type be a `Collection<CollegeMapsData>` or `List<CollegeMapsData>`? Returning a single object does not make sense to me. Unless you think this `CollegeMapsData` data structure itself will have a list for coordinates. In which case my question is; do you think it will contain anything but a list of coordinates/name pairs?

capstone/src/main/java/com/google/univiz/api/MapData.java
kaveh096: Ihsan's suggestion is not to use the exact interface name for return type (MapsData != MapData). I think he's saying the return type name should be similar (driven from) the interface name? For example the `MapData` api returns `MapsData` objects. I personally don't have a strong opinion.

Ihsan, remember that generally the burden of proof is on the code reviewer. If you're making a suggestion, you also need to provide the reasoning (preferably with citation) why you think this is the case.

capstone/src/main/java/com/google/univiz/api/MapData.java
mmurray22: hmm ok-so the search servlet would probably be able to provide that-a name + a uid?

capstone/src/main/java/com/google/univiz/api/MapData.java
mmurray22: yes indeed-I meant List<CollegeMapsData> and I accidentally changed it to be a single object. I will change it back.

capstone/src/main/java/com/google/univiz/api/MapData.java
mmurray22: done!

capstone/src/main/java/com/google/univiz/api/MapData.java
mmurray22: will do!

ccondit:
> I will make a doc responding to these concerns and share it with the group and then we can go from there.

Can you link the doc that you made into this PR for posterity?

capstone/src/main/java/com/google/univiz/api/MapData.java
```diff
@@ -1 +1,19 @@
 package com.google.univiz.api;
+
+/*
```
ccondit, line 3: nit - this public facing documentation should be javadoc.
```suggestion
/**
```

capstone/src/main/java/com/google/univiz/api/MapData.java
ccondit: Can you update the documentation to match the signature of the method?

capstone/src/main/java/com/google/univiz/api/MapData.java
```diff
@@ -1 +1,19 @@
...
+ * The MapData interface is responsible for all data management/processing specific for the maps
+ * feature of the UniViz webapp.
+ */
+public interface MapData {
+  /*
+   * Takes a list of college names and returns a CollegeMapsData object to the caller. This object
+   * contains all data relevant to the Maps feature.
+   *
+   * @args Takes a list of CollegeIDs, which is a class with two components: a college name and a
```
ccondit, line 12: I don't think `@args` is valid Javadoc. I think you want `@param`. Also, if you `@link` to the type involved you won't have to explain the type in this documentation. The advantage of that is that if the content of that type changes you won't have to update this comment.

capstone/src/main/java/com/google/univiz/api/CollegeId.java
```diff
@@ -0,0 +1,4 @@
+package com.google.univiz.api;
+
+/** */
```
ccondit, line 3: let's avoid empty javadoc, if possible...

capstone/src/main/java/com/google/univiz/api/MapData.java
ihsan314: That is true, I could have provided better evidence of what I was talking about.

I actually mistyped and meant `MapData`. And I also thought `CollegeMapsData` will be implementing `MapData`, so my thought process was `CollegeMapsData` should only be the return type if it offers something additional that returning `MapData` would not. Similar to how `List.subList` does not return `ArrayList` even though `ArrayList` would suffice in multiple scenarios.

I should have also mentioned it's more of a stylistic thing that isn't significant enough to block the PR on, but more of something for all of us to think about.

capstone/src/main/java/com/google/univiz/api/CollegeId.java
mmurray22: done! it's very simple, but I think it works for right now

capstone/src/main/java/com/google/univiz/api/CollegeMapsData.java
```diff
@@ -0,0 +1,6 @@
+package com.google.univiz.api;
+
+/** 
+ * The CollegeMapsData class will contain all college data that is specific to the Maps feature 
+ */
+public class CollegeMapsData {}
```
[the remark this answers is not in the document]
mmurray22: this data also has college admissions, college names, and other relevant information for the UniViz Maps feature, not just any Maps visualization. Should I clarify this in the Javadoc comments?

mmurray22:
Hi Chris! I forgot to ask about a comment you left in a previous PR because I didn't see what you were referencing. Seeing your repeat comment: "Looks like this still isn't updated." I now realize what the problem is-sorry for the repetition!

capstone/src/main/java/com/google/univiz/api/MapData.java
```diff
@@ -1 +1,19 @@
...
+public interface MapData {
+  /*
+   * Takes a list of college names and returns a CollegeMapsData object to the caller. This object
+   * contains all data relevant to the Maps feature.
+   *
+   * @param Takes a list of CollegeIDs, which is a class with two components: a college name and a 
+   * unique college id
+   *
+   * @return Returns a list of CollegeMapsData objects which contain all information necessary for
+   * a geographic visualization
```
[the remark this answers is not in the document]
mmurray22: updated that comment!

capstone/src/main/java/com/google/univiz/api/CollegeMapsData.java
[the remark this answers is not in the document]
ccondit: Avoid specific details in the class documentation. How will the admissions be used on the map?

ccondit:
> Hi Chris! I forgot to ask about a comment you left in a previous PR because I didn't see what you were referencing. Seeing your repeat comment: "Looks like this still isn't updated." I now realize what the problem is-sorry for the repetition!

no worries!

capstone/src/main/java/com/google/univiz/api/CollegeId.java
```diff
@@ -0,0 +1,7 @@
+package com.google.univiz.api;
+
+/**
+ * The CollegeId class will contain a college name and a unique college ID that can be used to 
+ * identify the college
```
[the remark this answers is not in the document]
mmurray22: I addressed it for CollegeId and CollegeMapsData

capstone/src/main/java/com/google/univiz/api/MapResource.java
```diff
@@ -1,7 +1,15 @@
...
+import java.util.List;
+
 /**
  * The MapResource interface is responsible for all data management/processing specific for the maps
  * feature of the UniViz webapp.
  */
-public interface MapResource {}
+public interface MapResource {
+  /**
+   * Takes a list of college names and returns a list of MapsData objects to the caller. This object
```
ccondit, line 11: Takes a list of CollegeIds?
```suggestion
   * Takes a list of college ids and returns a list of MapsData objects to the caller. This object
```

capstone/src/main/java/com/google/univiz/api/CollegeId.java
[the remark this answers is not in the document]
kaveh096: +1

capstone/src/main/java/com/google/univiz/api/CollegeId.java
```diff
@@ -0,0 +1,7 @@
+package com.google.univiz.api;
+
+/**
+ * The CollegeId class will contain a college name and a unique college ID that can be used to
+ * identify the college
```
kaveh096, line 5: Here is a suggestion to show you how you can document the behavior without actually leaking out implementation details.

```suggestion
 * This data structure contains all the college fields that are needed to uniquely identify a college. If two
 *  colleges have the same id, they they are the same college. Furthermore this structure may contain
 *  other alternative fields that could be used to uniquely identify a college.
```

capstone/src/main/java/com/google/univiz/api/CollegeId.java
```diff
@@ -0,0 +1,7 @@
...
+ */
+public class CollegeId {}
```
kaveh096, line 7: I have a feeling this would later become an AutoValue and that's why you didn't make it `final. But if this is not the case then mark it as `final`.

```suggestion
public final class CollegeId {}
```

capstone/src/main/java/com/google/univiz/api/MapsData.java
```diff
@@ -0,0 +1,4 @@
+package com.google.univiz.api;
+
+/** The MapsData class will contain all college data that is specific to the Maps feature */
+public class MapsData {}
```
kaveh096, line 4: Could this also be `final` or will be an `AutoValue`?

capstone/src/main/java/com/google/univiz/api/CollegeId.java
mmurray22: I do believe this will eventually become Autovalue :)

capstone/src/main/java/com/google/univiz/api/MapsData.java
mmurray22: AutoValue :)

PR #48, opened 2020-08-03
ScorecardTest.java implementation

biancamacias:
This is a test to assert that the JSON parsing is working. It is unfinished, as I haven't added the sample JSON, but this is a simple test I can do once I'm finished with the ```ScorecardData.java``` class.

### DIFF
--- a/capstone/src/test/resources/com/google/univiz/scorecard/ScorecardTest.java
+++ b/capstone/src/test/resources/com/google/univiz/scorecard/ScorecardTest.java
@@ -1,0 +1,18 @@
+package com.google.univiz.scorecard;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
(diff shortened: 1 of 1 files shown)


capstone/src/test/resources/com/google/univiz/scorecard/ScorecardTest.java
```diff
@@ -0,0 +1,19 @@
...
+public final class ScorecardTest {
+
+  @Test
+  public void parseJsonTest() {
+    // TODO(biancamacias): insert sample JSON
+    String json = "{}";
+    ScorecardData college = new Gson().fromJson(json, ScorecardData.class);
+
+    asserThat(college.name().isEqualTo("New York University"));
+    asserThat(college.city().isEqualTo("New York City"));
```
ccondit, line 17: I don't believe this PR will build. Can you sync past #47 so that the Java builder check runs on this PR?

PR #50, opened 2020-08-04
Autocomplete JavaScript: Part 1

mmurray22:
Adds a JavaScript implementation of autocomplete. This is just the bare bones JS gotten from the w3 website which gives enough view of the potential functionality. Currently using dummy data. 

Here is a screenshot of the current functionality:
univiz_autocomplete

Note on JS Linter: Not sure how to/if I should resolve remaining issues.


capstone/src/main/webapp/autocomplete.js
```diff
@@ -0,0 +1,89 @@
...
+    let a = this.value;
+    let b = this.value;
+    let i = this.value;
+    let val = this.value;
+    closeAllLists();
+    if (!val) { 
+        return false;
+    }
+    currentFocus = -1;
+    a = document.createElement('DIV');
```
ihsan314, line 21: Instead of assigning `this.value` to `a`, you can make `a` `const` here since you don't assign elsewhere.

capstone/src/main/webapp/autocomplete.js
```diff
@@ -0,0 +1,89 @@
...
+    a.setAttribute('id', this.id + 'autocomplete-list');
+    a.setAttribute('class', 'autocomplete-items');
+    this.parentNode.appendChild(a);
+    for (i = 0; i < arr.length; i++) {
+      if (arr[i].substr(0, val.length).toUpperCase() == val.toUpperCase()) {
+        b = document.createElement('DIV');
```
ihsan314, line 27: Ditto.

capstone/src/main/webapp/autocomplete.js
```diff
@@ -0,0 +1,89 @@
...
+ * Link: https://www.w3schools.com/howto/howto_js_autocomplete.asp
+*/
+function autocomplete(inp, arr) {
+  let currentFocus;
+  inp.addEventListener('input', function(e) {
+    let a = this.value;
+    let b = this.value;
+    let i = this.value;
+    let val = this.value;
+    closeAllLists();
```
ihsan314, line 16: Are you missing a parameter here? If the method signature is right and this function call is correct, it feels off to call a function that is defined to take a parameter without one. See if you can pass in an empty list instead.

capstone/src/main/webapp/autocomplete.js
```diff
@@ -0,0 +1,89 @@
...
+    a = document.createElement('DIV');
+    a.setAttribute('id', this.id + 'autocomplete-list');
+    a.setAttribute('class', 'autocomplete-items');
+    this.parentNode.appendChild(a);
+    for (i = 0; i < arr.length; i++) {
+      if (arr[i].substr(0, val.length).toUpperCase() == val.toUpperCase()) {
+        b = document.createElement('DIV');
+        b.innerHTML = '<strong>' + arr[i].substr(0, val.length) + '</strong>';
+        b.innerHTML += arr[i].substr(val.length);
+        b.innerHTML += '<input type='+'hidden'+'value='+arr[i]+'>';
```
ihsan314, line 30: Could possibly use the DOM instead.

biancamacias:
> Adds a JavaScript implementation of autocomplete. This is just the bare bones JS gotten from the w3 website which gives enough view of the potential functionality. Currently using dummy data.
> Here is a screenshot of the current functionality:
> image
> 
> Not passing all the formatting tests, but working on it

so cool!

capstone/src/main/webapp/autocomplete.js
mmurray22: thanks! will do!

capstone/src/main/webapp/autocomplete.js
mmurray22: thanks! will do!

capstone/src/main/webapp/autocomplete.js
mmurray22: will look into that :)

mmurray22:
Alrighty I've made a lot of changes here, and I hope it's looking better! Added a test file without a test yet, but wanted to see how I was doing so far. Thanks!

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,112 @@
...
+      return false;
+    }
+    currentFocus = -1;
+    const autocompleteList = document.createElement('DIV');
+    autocompleteList.setAttribute('id', this.id + 'autocomplete-list');
+    autcompleteList.setAttribute('class', 'autocomplete-items');
+    this.parentNode.appendChild(autocompleteList);
+    for (let i = 0; i < arr.length; i++) {
+      if (arr[i].substr(0, val.length).toUpperCase() == val.toUpperCase()) {
+        const listElmt = document.createElement('DIV');
```
ihsan314, line 28: Ditto.

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,112 @@
...
+  let currentFocus = -1;
+  /* Event occurrance when input is provided to autocomplete field */
+  inp.addEventListener('input', function(e) {
+    closeAllLists();
+    const val = this.value;
+    if (!val) {
+      return false;
+    }
+    currentFocus = -1;
+    const autocompleteList = document.createElement('DIV');
```
ihsan314, line 22: Make `'DIV'` lowercase for consistency with how you would actually write HTML.

mmurray22: done!

capstone/src/main/webapp/js/autocomplete.js
mmurray22: ditto to which comment?

capstone/src/main/webapp/index.html
```diff
@@ -3,13 +3,16 @@
...
     <h2>Welcome to UniViz!</h2>
     <form autocomplete="off">
       <label for="search">Search for colleges:</label><br>
-      <input type="text" id="search" name="search"><br>
+      <input type="text" id="search" name="search" placeholder="College Name"><br>
       <input type="submit" value="See your colleges!">
     </form>
   </body>
+  <script src="js/autocomplete.js"></script>
+  <script src="js/autocompleteTest.js"></script>
```
ihsan314, line 17:
```suggestion
```
Unless there's some function in `autocompleteTest.js` that you need inside this file.

capstone/src/main/webapp/autocomplete.js
kaveh096: +1 to DOM createElement

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,140 @@
+const inp = document.getElementById('search');
+const currentFocus = -1;
+/**
+ * @param {Object} autocompleteListElmt
```
ccondit, line 4: Have look at [section 7](https://google.github.io/styleguide/jsguide.html#jsdoc) in the JSDoc style guide and see if you can bring the JSDoc in this file more in line with the style guide...

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,140 @@
...
+  if (!autocompleteListElmt) {
+    return false;
+  }
+  removeActiveTag(autocompleteListElmt);
+  if (currentFocus >= autocompleteListElmt.length) {
+    currentFocus = 0;
+  } else if (currentFocus < 0) {
+    currentFocus = (autocompleteListElmt.length - 1);
+  }
+  autocompleteListElmt[currentFocus].classList.add('autocomplete-active');
```
ccondit, line 18: Seems like `autocomplete-active` is a good candidate for a constant!

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,140 @@
...
+/* Event occurrance when input is provided to autocomplete field */
+document.addEventListener('input', function(e) {
+  closeAllLists();
+  const val = this.value;
+  if (!val) {
+    return false;
+  }
+  const arr = getListOfSuggestions();
+  currentFocus = -1;
+  const autocompleteList = document.createElement('div');
```
ccondit, line 107: can this element be defined in the HTML?

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,140 @@
...
+  autocompleteList.setAttribute('id', this.id + 'autocomplete-list');
+  autcompleteList.setAttribute('class', 'autocomplete-items');
```
ccondit, line 109: Does this code work?
```suggestion
  autocompleteList.setAttribute('class', 'autocomplete-items');
```

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,140 @@
...
+ */
+function getRelevantDataSuggestions(arr, val) {
+  const trimArr = [];
+  for (const arrElt of arr) {
+    if (arrElt.substr(0, val.length).toUpperCase() == val.toUpperCase()) {
+      trimArr.push(arrElt);
+    }
+  }
+  return trimArr;
+}
```
ccondit, line 72: OOC - shouldn't the server be making sure that relevant suggests are returned to the client?

We could also imagine a case where a client sends in "UCLA" but the response is the entire name. Wouldn't we end up filtering out such suggestions here?

kaveh096: I think this is the fake server implementation so we can manually testing autocomplete.

kaveh096: Actually i don't know. Cause getListOfSuggestions also looks like it could be the fake server.

capstone/src/test/javascript/search/autocompleteSpec.js
```diff
@@ -0,0 +1,8 @@
+describe('Autocomplete Suggests correct array', () => {
```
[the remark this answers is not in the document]
mmurray22: yes :) I was planning on expanding on the test suite in a new PR, but I can add more tests to this PR :)

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,140 @@
...
+
+/**
+ * @param {Object} autocompleteListElmt
+ * @return {void}
+ * Removes the 'active' tag from an autocomplete elmt
+ */
+function removeActiveTag(autocompleteListElmt) {
+  for (listElmt in autocompleteListElmt) {
+    if (true) {
```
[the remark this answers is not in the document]
mmurray22: yes it will be! However, the JS linter complained when I didn't have it, thus why it is there. Is there a better way to address this problem?

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,140 @@
...
+ * @return {bool}
+ * Adds the 'active' tag to an autocomplete elmt
+ */
+function addActiveTag(autocompleteListElmt) {
+  if (!autocompleteListElmt) {
+    return false;
+  }
+  removeActiveTag(autocompleteListElmt);
+  if (currentFocus >= autocompleteListElmt.length) {
+    currentFocus = 0;
```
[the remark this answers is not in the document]
mmurray22: I'm currently testing that. However, I will not that the JS linter would fail if the currentFocus was anything other than a const

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,140 @@
...
+      listElmt.classList.remove('autocomplete-active');
+    }
+  }
+}
+
+/**
+ * @param {Object} elmnt
+ * @param {Object} inp
+ * @return {void}
+ * Closes dropdown autocomplete list(s)
```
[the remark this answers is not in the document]
mmurray22: I guess in practice there will only be one list! Thanks for pointing that out!

univiz.conf.js
```diff
@@ -15,8 +15,8 @@ module.exports = function(config) {
 
     // list of files / patterns to load in the browser
     files: [
-      'capstone/src/test/javascript/**/*[Ss]pec.js',
       'capstone/src/main/webapp/js/*.js',
+      'capstone/src/test/javascript/**/*[Ss]pec.js',
```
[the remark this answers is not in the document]
mmurray22: I fixed the order when I was debugging the problems with the test. I will move them back, and if it still works, I'll remove the change 👍

capstone/src/main/webapp/js/autocomplete.js
mmurray22: will do!

capstone/src/test/javascript/search/autocompleteSpec.js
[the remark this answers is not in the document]
mmurray22: just added another test!

capstone/src/main/webapp/js/autocomplete.js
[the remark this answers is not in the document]
mmurray22: used lambda!

capstone/src/main/webapp/js/autocomplete.js
mmurray22: Yeah this is like a fake server implementation so we can test the autocomplete functionality.

mmurray22: Is there a better way to approach this?

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,140 @@
...
+ */
+function closeAllLists(elmnt, inp) {
+  const autoItems = document.getElementsByClassName('autocomplete-items');
+  for (item in autoItems) {
+    if (elmnt != item && elmnt != inp) {
+      item.parentNode.removeChild(item);
+    }
+  }
```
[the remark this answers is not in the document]
mmurray22: done!

capstone/src/main/webapp/js/autocomplete.js
mmurray22: It could-however, since the id and the class are both added in the program, I'm not sure how to uniquely identify the autocompleteList in order to call it in the javascript function?

capstone/src/main/webapp/js/autocomplete.js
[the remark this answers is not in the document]
mmurray22: It seems that currentFocus cannot be a constant in order to get the desired functionality. However, this means that the linter is complaining and thus failing it's tests...any advice on how to approach this?

capstone/src/main/webapp/js/autocomplete.js
mmurray22: Looked at the section 7 documentation & updated a lot of the JSDoc-let me know if you expected more to change!

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,149 @@
...
+  autocompleteList.setAttribute('id', searchInput.id + listId);
+  autocompleteList.setAttribute('class', itemClass);
+  searchInput.parentNode.appendChild(autocompleteList);
+  const trimArr = getRelevantDataSuggestions(arr, val);
+  displaySuggestions(trimArr, autocompleteList, val);
+}
+
+/**
+ * Event occurance when arrow keys are pressed 
+ * @param {event} e
```
fabielzuniga, line 121: What is the type of the Event?

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,149 @@
...
+ * @return {void}
+ */ 
+function keyDown(e) {
+  console.log("Key down!");
+  let listElmt = document.getElementById(searchInput.id + listId);
+  if (listElmt) {
+    listElmt = listElmt.getElementsByTagName('div');
+  }
+  if (e.keyCode == 40) {/* Up key */
```
fabielzuniga, line 130: 40, 38 and 13 could be defined as constants and then the comment describing what the literal represents wouldn't be needed. Thoughts?

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,149 @@
...
+    if (selectedSuggElt > -1) {
+      if (listElmt) {
+        listElmt[selectedSuggElt].click();
+      }
+    }
+  }
+}
+
+/* Event occurance when mouse is clicked */
+document.addEventListener('click', function(e) {
```
fabielzuniga, line 147: Google favors the arrow function

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,149 @@
+const searchInput = document.getElementById('search');
+let selectedSuggElt = -1;
+const itemClass = 'autocomplete-items';
+const activeClass = 'autocomplete-active';
+const listId = 'autocomplete-list';
```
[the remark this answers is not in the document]
mmurray22: done!

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,149 @@
...
+  for (const item of autoItems) {
+    if (item != elmnt && item != searchInput) {
+      item.parentNode.removeChild(item);
+    }
+  }
+}
+
+/**
+ * Current placeholder until servlet is created
+ * @return{Object} arr
```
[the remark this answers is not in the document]
mmurray22: done!

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,149 @@
+const searchInput = document.getElementById('search');
+let selectedSuggElt = -1;
+const itemClass = 'autocomplete-items';
+const activeClass = 'autocomplete-active';
+const listId = 'autocomplete-list';
+/**
+ * Adds the 'active' tag to an autocomplete elmt
+ * @param {Object} autocompleteListElmt
```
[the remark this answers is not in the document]
mmurray22: I'm not sure exactly how to describe it...should I just say it's a div?

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,149 @@
...
+  } else if (selectedSuggElt < 0) {
+    selectedSuggElt = (autocompleteListElmt.length - 1);
+  }
+  autocompleteListElmt[selectedSuggElt].classList.add(activeClass);
+}
+
+/**
+ * Removes the 'active' tag from an autocomplete elmt
+ * @param {Object} autocompleteListElmt
+ * @return {void}
```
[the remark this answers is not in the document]
mmurray22: removed!

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,149 @@
...
+ */
+function removeActiveTag(autocompleteListElmt) {
+  for (elmt of autocompleteListElmt) {
+    elmt.classList.remove(activeClass);
+  }
+}
+
+/**
+ * Closes dropdown autocomplete list
```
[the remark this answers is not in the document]
mmurray22: done!

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,149 @@
+const searchInput = document.getElementById('search');
+let selectedSuggElt = -1;
```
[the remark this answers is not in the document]
mmurray22: ok, I updated the name to SELECTED_ELMT_POS-does this seem more apparent to the reader?

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,149 @@
...
+  }
+}
+
+/**
+ * Closes dropdown autocomplete list
+ * @param {Object} elmnt
+ * @param {Object} inp
+ * @return {void}
+ */
+function closeAllLists(elmnt) {
```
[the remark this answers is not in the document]
mmurray22: done!

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,149 @@
...
+
+/**
+ * Takes relevant suggestions and displays them in DOM
+ * @param {Object} trimArr
+ * @param {Object} autocompleteList
+ * @return {void}
+ */
+function displaySuggestions(trimArr, autocompleteList, val) {
+  for (arrElt of trimArr) {
+    if (true) {
```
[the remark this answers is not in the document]
mmurray22: no :) updated!

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,149 @@
...
+      const listElmt = document.createElement('DIV');
+      listElmt.innerHTML = '<strong>' +
+                           arrElt.substr(0, val.length) +
+                           '</strong>';
+      listElmt.innerHTML += arrElt.substr(val.length);
+      listElmt.innerHTML += '<input type=\'hidden\' value=\'arrElt\'>';
```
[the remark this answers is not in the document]
mmurray22: done!

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,149 @@
...
+      listElmt.addEventListener('click', function(e) {
```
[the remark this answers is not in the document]
mmurray22: done!

capstone/src/main/webapp/js/autocomplete.js
mmurray22: keypress! updated!

capstone/src/main/webapp/js/autocomplete.js
mmurray22: yes! I can make the comments the constant names essentially and no comments necessary :) thank you!

capstone/src/main/webapp/js/autocomplete.js
mmurray22: done!

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,149 @@
...
+ * Takes relevant suggestions and displays them in DOM
+ * @param {Object} trimArr
+ * @param {Object} autocompleteList
+ * @return {void}
+ */
+function displaySuggestions(trimArr, autocompleteList, val) {
+  for (arrElt of trimArr) {
+    if (true) {
+      const listElmt = document.createElement('DIV');
+      listElmt.innerHTML = '<strong>' +
```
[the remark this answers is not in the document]
mmurray22: Ah thanks for the suggestion! Do you have a link/suggestions on where to look for google.dom? I tried looking it up in general, but I didn't see much helpful documentation. Thanks in advance!

capstone/src/main/webapp/js/autocomplete.js
mmurray22: ah ok...would you recommend doing that now or should I change that in a follow-up PR that will actually get the data from the servlet and connect wires?

capstone/src/main/webapp/js/autocomplete.js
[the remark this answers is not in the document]
mmurray22: well it doesn't like the lambda, but I still got rid of the true statement!

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,140 @@
...
+  autocompleteListElmt[currentFocus].classList.add('autocomplete-active');
+}
+
+/**
+ * @param {Object} autocompleteListElmt
+ * @return {void}
+ * Removes the 'active' tag from an autocomplete elmt
+ */
+function removeActiveTag(autocompleteListElmt) {
+  for (listElmt in autocompleteListElmt) {
```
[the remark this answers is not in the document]
mmurray22: it did not seem to like that :( got loads of errors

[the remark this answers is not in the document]
mmurray22: Same thing goes for any other lambdas I tried to use-is it necessary to use a lambda here or should that be a stretch goal?

capstone/src/main/webapp/js/autocomplete.js
mmurray22: sounds good! doing that now :)

mmurray22:
Note: I am failing two lint tests because I am not using two functions within the javascript file, I'm using them in the HTML file. Any advice on this front is welcome!

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,148 @@
...
+ * Event occurrance when input is provided to autocomplete field
+ * @return {void}
+ */
+function giveSuggestions() {
+  closeAllElmntExcept();
+  const val = SEARCH_INPUT.value;
+  if (!val) {
+    return false;
+  }
+  const arr = getListOfSuggestions();
```
kaveh096, line 111: So i was thinking about the async problem @ccondit brought up in the standup. One solution that came to my mind was to check if input data is still the same value after the fetch completed. Basically right here you can check if `SEARCH_INPUT.value == val`. IF this check failed, it means there is a new query and the suggestions for the old one can be discarded right here. Just a suggestion. I'm not sure how it would that affect the overall auto-complete flow.

capstone/src/main/webapp/js/autocomplete.js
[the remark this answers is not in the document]
ihsan314: I think the appropriate type is an [`HTMLDivElement`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLDivElement).

capstone/src/main/webapp/js/autocomplete.js
mmurray22: I tried this approach kaveh and it doesn't seem to work :/ I am thinking of ways to address this problem in the following PR (part 2)

ihsan314:
> Note: I am failing two lint tests because I am not using two functions within the javascript file, I'm using them in the HTML file. Any advice on this front is welcome!

[Here](https://eslint.org/docs/rules/no-unused-vars#varsignorepattern) is how to ignore the no-unused-var rule selectively.

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,167 @@
...
+  const listElt = document.createElement('li');
+  listElt.innerHTML = validSuggestion;
+  parent.appendChild(listElt);
+}
+
+/**
+ * Current placeholder until servlet is created
+ * @return{Array<string>} arr
+ */
+function getListOfSuggestions() {
```
kaveh096, line 71: In reality this would be an async method. So make sure your rendering logic still works with a real server too. No actions here. Just a heads up.

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,167 @@
...
+ * @param {Object} trimArr
+ * @param {Object} autocompleteList
+ * @param {string} val
+ * @return {void}
+ */
+function displaySuggestions(trimArr, autocompleteList, val) {
+  for (arrElt of trimArr) {
+    const listElmt = document.createElement('DIV');
+    listElmt.innerHTML = '<strong>' +
+                         arrElt.substr(0, val.length) +
```
kaveh096, line 104: You are assuming that the search query is always at the beginning of the array element. Are we sure that this is always the case? If not you can use [String.indexOf](https://www.w3schools.com/jsref/jsref_indexof.asp) to be safe.

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,167 @@
...
+  return arr;
+}
+
+/**
+ * Identifies and returns relevant suggestions in the arr
+ * @param {string[]} arr
+ * @param {string} val
+ * @return {string[]} trimArr
+ */
+function getRelevantDataSuggestions(arr, val) {
```
kaveh096, line 82: Why do we need to trim the array based on whether the result starts with the val (search input)? Do we expect scorecard to return results that do not pass this check? If yes, then why do we need to drop them here? For example if someone searched for `tanford`, should this not match `Stanford`?

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,167 @@
...
+}
+
+/**
+ * Event occurance when arrow keys are pressed
+ * @param {keypress} e
+ * @return {void}
+ */
+function keyDown(e) {
+  let listElmt = document.getElementById(LIST_ID);
+  if (listElmt) {
```
ccondit, line 145: OOC - what happens to the remainder of the method if listElmnt is false?

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,167 @@
...
+    listElmt = listElmt.getElementsByTagName('div');
+  }
+  if (e.keyCode == UP_KEY) {
+    selectedElmntPos++;
+    addActiveTag(listElmt);
+  } else if (e.keyCode == DOWN_KEY) {
```
ccondit, line 151: [Should these equality checks be ===](https://google.github.io/styleguide/jsguide.html#features-equality-checks)?

capstone/src/main/webapp/autocomplete.js
ccondit: Did this get addressed?

capstone/src/main/webapp/index.html
```diff
@@ -3,13 +3,18 @@
   <head>
     <title>UniViz Main Page</title>
     <meta charset="utf-8">
+    <link rel="stylesheet" href="style.css">
+    <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.5.1/jquery.min.js" type="text/javascript"></script>
```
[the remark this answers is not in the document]
mmurray22: that is an old artifact :) removed!

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,167 @@
...
+  SEARCH_INPUT.parentNode.appendChild(autocompleteList);
+  const trimArr = getRelevantDataSuggestions(arr, val);
+  displaySuggestions(trimArr, autocompleteList, val);
+  return true;
+}
+
+/**
+ * Event occurance when arrow keys are pressed
+ * @param {keypress} e
+ * @return {void}
```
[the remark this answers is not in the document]
mmurray22: updated!

capstone/src/main/webapp/js/autocomplete.js
mmurray22: yes!

capstone/src/main/webapp/style.css
```diff
@@ -2,3 +2,46 @@ body {
...
+  display: inline-block;
+  background-color: dodgerblue;
+  color: #fff;
+}
+
+.autocomplete-items div {
+  padding: 10px;
+  cursor: pointer;
+  background-color: #fff;
+  border-bottom: 1px solid #d4d4d4;
```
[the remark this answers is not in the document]
mmurray22: I alphabetized all the style tags that were from the w3 schools website

capstone/src/main/webapp/js/autocomplete.js
mmurray22: Hmmm I was assuming this was the case. Like you'd only get 'Stanford' as a suggestion if you look up 'Stan' and not if you look up 'tanford'. Is that ok or should I look into an alternative way?

capstone/src/main/webapp/js/autocomplete.js
mmurray22: See my reply to your previous comment: 
Hmmm I was assuming this was the case. Like you'd only get 'Stanford' as a suggestion if you look up 'Stan' and not if you look up 'tanford'. Is that ok or should I look into an alternative way?

capstone/src/main/webapp/autocomplete.js
mmurray22: I ended up not using DOM :)

capstone/src/main/webapp/js/autocomplete.js
mmurray22: This was discussed in standup :)

capstone/src/main/webapp/js/autocomplete.js
mmurray22: Resolved in standup :)

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,167 @@
...
+    }
+  }
+}
+
+/**
+ * Adds valid suggestions to list of stored suggestions
+ * @param{string} validSuggestion
+ */
+function keepTrackOfSuggestions(validSuggestion) {
+  const parent = document.getElementById('suggestions');
```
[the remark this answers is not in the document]
mmurray22: I tried that :/ unfortunately it seems to work better to define in the JS. In a later cleanup of the JS I will attempt this again!

capstone/src/main/webapp/js/autocomplete.js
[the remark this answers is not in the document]
mmurray22: done!

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,167 @@
...
+      if (listElmt) {
+        listElmt[selectedElmntPos].click();
+      }
+    }
+  }
+}
+
+/* Event occurance when mouse is clicked */
+document.addEventListener('click', (e) => {
+  closeAllElmntExcept(e.target, SEARCH_INPUT);
```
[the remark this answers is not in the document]
mmurray22: done!

capstone/src/main/webapp/js/autocomplete.js
mmurray22: added the suggestion btw and it works fine now :) it'll be telling to see how it updates in the next pr

capstone/src/main/webapp/js/autocomplete.js
mmurray22: it will eventually exit out, but I added an early return :)

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,167 @@
...
+ */
+function giveSuggestions() {
+  closeAllElmntExcept();
+  const val = SEARCH_INPUT.value;
+  if (!val) {
+    return false;
+  }
+  const arr = getListOfSuggestions();
+  selectedElmntPos = -1;
+  const autocompleteList = document.createElement('DIV');
```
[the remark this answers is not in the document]
mmurray22: made it consistent :)

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,167 @@
...
+
+/**
+ * Adds valid suggestions to list of stored suggestions
+ * @param{string} validSuggestion
+ */
+function keepTrackOfSuggestions(validSuggestion) {
+  const parent = document.getElementById('suggestions');
+  const listElt = document.createElement('li');
+  listElt.innerHTML = validSuggestion;
+  parent.appendChild(listElt);
```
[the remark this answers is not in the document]
mmurray22: Hmm that is interesting-should I have an internal data structure to keep track of it?

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,167 @@
...
+ * Event occurance when arrow keys are pressed
+ * @param {keypress} e
+ * @return {void}
+ */
+function keyDown(e) {
+  let listElmt = document.getElementById(LIST_ID);
+  if (listElmt) {
+    listElmt = listElmt.getElementsByTagName('div');
+  }
+  if (e.keyCode == UP_KEY) {
```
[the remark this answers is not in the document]
mmurray22: done!

capstone/src/main/webapp/style.css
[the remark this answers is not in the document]
ccondit: looks like the styles got reordered, but the declarations are still not alphabetized. mind updating that?

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,169 @@
...
+/*
+ * Largely inspired by JS code from W3Schools: https://www.w3schools.com/howto/howto_js_autocomplete.asp
+ */
+const SEARCH_INPUT = document.getElementById('search');
+const ITEM_CLASS = 'autocomplete-items';
+const ACTIVE_CLASS = 'autocomplete-active';
+const LIST_ID = 'autocomplete-list';
+const UP_KEY = 'ArrowUp';
+const DOWN_KEY = 'ArrowDown';
+const ENTER = 'Enter';
```
ccondit, line 12: nit - match the names of the other key related constants?
```suggestion
const ENTER_KEY = 'Enter';
```

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,169 @@
...
+let selectedElmntPos = -1;
```
ccondit, line 13: Consider renaming this variable so the intent is more clear:
selectedSuggestionPosition? something along those lines...

capstone/src/main/webapp/js/autocomplete.js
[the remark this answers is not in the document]
ccondit: Didn't you mention offline that you were going to store the ID of the college as the ID of the list element? Can you check if if the list already has an element with that ID before adding it again?

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,169 @@
...
+ */
+function giveSuggestions() {
+  closeAllElmntExcept();
+  const val = SEARCH_INPUT.value;
+  if (!val) {
+    return;
+  }
+  if (SEARCH_INPUT.value !== val) {
+    return;
+  }
```
ccondit, line 127: What's the purpose of this condition? To the casual reader it appears that it would never return since val was just assigned to SEARCH_INPUT.value.

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,169 @@
...
+  const arr = getListOfSuggestions();
+  selectedElmntPos = -1;
+  const autocompleteList = document.createElement('div');
+  autocompleteList.setAttribute('id', LIST_ID);
+  autocompleteList.setAttribute('class', ITEM_CLASS);
+  SEARCH_INPUT.parentNode.appendChild(autocompleteList);
+  const trimArr = getRelevantDataSuggestions(arr, val);
```
ccondit, line 134: what does `trim` mean in this context?

capstone/src/main/webapp/js/autocomplete.js
mmurray22: Kaveh suggested addition in an earlier comment. It did not break anything so I decided to keep it in. I can remove though.

capstone/src/main/webapp/js/autocomplete.js
mmurray22: done!

mmurray22:
@ccondit  Sounds good (Referencing your comment here https://github.com/googleinterns/univiz/pull/50#discussion_r478576329) I will address this problem in Autocomplete Part 2

capstone/src/main/webapp/js/autocomplete.js
mmurray22: changed the var name to relevantSuggestions

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,166 @@
...
+  } else {
+    return;
+  }
+  if (e.code === DOWN_KEY) {
+    selectedSuggestionPosition++;
+    addActiveTag(listElmt);
+  } else if (e.code === UP_KEY) {
+    selectedSuggestionPosition--;
+    addActiveTag(listElmt);
+  } else if (e.code === ENTER) {
```
ccondit, line 152: Doesn't this need to change to ENTER_KEY?

capstone/src/main/webapp/style.css
```diff
@@ -2,3 +2,46 @@ body {
   background-color: white;
   margin: 0;
 }
+
+/*
+ * Code source from W3Schools: https://www.w3schools.com/howto/howto_js_autocomplete.asp
+ */
+.autocomplete {
+  position: relative;
```
ccondit, line 10: Looks like this didn't get addressed: https://github.com/googleinterns/univiz/pull/50#discussion_r478573816

capstone/src/main/webapp/js/autocomplete.js
ccondit: Trying to grok this with [this comment](https://github.com/googleinterns/univiz/pull/50#discussion_r478592139):

@mmurray22 - looks like it did get removed?

also look like @kaveh096 is suggesting adding this *after* the fetch call is completed, not before...

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,166 @@
...
+
+/**
+ * Closes dropdown autocomplete list
+ * Does not close the element provided
+ * @param {Object} elmnt
+ */
+function closeAllElmntExcept(elmnt) {
+  const autoItems = document.getElementsByClassName(ITEM_CLASS);
+  for (const item of autoItems) {
+    if (item != elmnt && item != SEARCH_INPUT) {
```
ccondit, line 50: Would `SEARCH_INPUT` ever have the `ITEM_CLASS` class?

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,166 @@
...
+      item.parentNode.removeChild(item);
+    }
+  }
+}
+
+/**
+ * Adds valid suggestions to list of stored suggestions
+ * @param{string} validSuggestion
```
ccondit, line 58: Can you have a look at the JSDoc throughout - I think there should be a space between the `@param` and the `{` - I think I saw that in a couple of other spots, as well.

capstone/src/main/webapp/js/autocomplete.js
kaveh096: My suggestion was to return early without updating the list element in order to prevent stale data from being shown to user. This was meant to be a basic implementation to prevent race conditions. So we do `val = SEARCH_INPUT.value` before fetch and once the fetch completed, we add this if statement. Not sure why i didn't work. But in any case, feel free to remove it and ignore my solution if it's not working.

capstone/src/main/webapp/index.html
```diff
@@ -3,13 +3,17 @@
...
   <body>
     <h2>Welcome to UniViz!</h2>
     <form autocomplete="off">
       <label for="search">Search for colleges:</label><br>
-      <input type="text" id="search" name="search"><br>
+      <input type="text" id="search" name="search" placeholder="College Name" oninput="giveSuggestions()" onkeydown="keyDown(event)"><br>
       <input type="submit" value="See your colleges!">
     </form>
+    <p>Current List of Suggested Colleges:</p>
+    <ul id="suggestions"></ul>
```
kaveh096, line 16: First i thought this is the list item that is showing the suggestions. But after reading the code i realized this is the list of colleges that the user selected. So a better name / text would help. Something like "Current list of selected colleges". Same with element id.

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,166 @@
+/* eslint-disable no-unused-vars */
+
+/*
+ * Largely inspired by JS code from W3Schools: https://www.w3schools.com/howto/howto_js_autocomplete.asp
+ */
+const SEARCH_INPUT = document.getElementById('search');
+const ITEM_CLASS = 'autocomplete-items';
+const ACTIVE_CLASS = 'autocomplete-active';
+const LIST_ID = 'autocomplete-list';
```
kaveh096, line 9: There are two lists in this file the list of auto-complete suggestions and the list of selected colleges. Use a more specific variable name to clarify which one this is.

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,166 @@
...
+      closeAllElmntExcept();
+    });
+    autocompleteList.appendChild(listElmt);
+  }
+}
+
+/**
+ * Event occurrance when input is provided to autocomplete field
+ */
+function giveSuggestions() {
```
kaveh096, line 119: nit: in term of method naming, giveSuggestions sounds like it's going to give the caller (return to caller) new suggestions. But in reality it's getting suggestions and populating them on UI. So something like {display,populate,show}Suggestions might read better. That being said you already have a `displaySuggestions` method that is doing that actual displaying. So this method feels like pre-work.

Here are few other name suggestions: triggerAutocomplete, refreshSuggestions, etc.

Alternatively you can rename the current displaySuggestions method to something else like populateSuggestionsInList. Then rename this method to displaySuggestions.

capstone/src/main/webapp/js/autocomplete.js
kaveh096: +1 to element being defined in html. Fundamentally we have to lists in this html, a list of suggestions, and a list of selected colleges. having them both in html helps the reader a lot.

capstone/src/main/webapp/js/autocomplete.js
mmurray22: changed!

capstone/src/main/webapp/js/autocomplete.js
mmurray22: made consistent!

capstone/src/main/webapp/style.css
mmurray22: sorry-didn't see the new comments! I hope it is addressed now!

capstone/src/main/webapp/js/autocomplete.js
mmurray22: nope! removed :)

mmurray22:
https://github.com/googleinterns/univiz/pull/50#discussion_r478731581

@ccondit Have I addressed this comment fully enough?

capstone/src/main/webapp/index.html
mmurray22: done! updated the label

capstone/src/main/webapp/js/autocomplete.js
mmurray22: done!

capstone/src/main/webapp/js/autocomplete.js
mmurray22: I renamed the giveSuggestions to getAllProposedSuggestions-> how does that work?

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,166 @@
...
+const UP_KEY = 'ArrowUp';
+const DOWN_KEY = 'ArrowDown';
+const ENTER_KEY = 'Enter';
+let selectedSuggestionPosition = -1;
+
+/**
+ * Adds the 'active' tag to an autocomplete elmt
+ * @param {HTMLDivElement} autocompleteListElmt
+ */
+function addActiveTag(autocompleteListElmt) {
```
fabielzuniga, line 19: If possible try to avoid abbreviations. autocompleteListElement would read better.

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,166 @@
...
+ * @return {Array<string>} arr
+ */
+function getListOfSuggestions() {
+  const arr = ['Hallo', 'Hello', 'Hi', 'Hiya', 'Howdy', 'Wassup'];
+  return arr;
+}
+
+/**
+ * Identifies and returns relevant suggestions in the arr
+ * @param {string[]} arr
```
fabielzuniga, line 78: This should be `Array<string>` instead of string[]
Same comment below

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,166 @@
...
+ * @param {string} val
+ */
+function displaySuggestions(relevantSuggestions, autocompleteList, val) {
+  for (arrElt of relevantSuggestions) {
+    const listElmt = document.createElement('div');
+    listElmt.innerHTML = '<strong>' +
+                         arrElt.substr(0, val.length) +
+                         '</strong>';
+    listElmt.innerHTML += arrElt.substr(val.length);
+    const cpyArrElt = arrElt;
```
fabielzuniga, line 106: Same comment about abbreviation.

Same comment for all instances in this review.

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,166 @@
...
+ * Identifies and returns relevant suggestions in the arr
+ * @param {string[]} arr
+ * @param {string} val
+ * @return {string[]} relevantSuggestions
+ */
+function getRelevantDataSuggestions(arr, val) {
+  const relevantSuggestions = [];
+  val = val.toUpperCase();
+  for (const arrElt of arr) {
+    if (arrElt.substr(0, val.length).toUpperCase() === val.toUpperCase()) {
```
fabielzuniga, line 86: Wouldn't startWith method be equivalent?
`arrayElement.toUpperCase().startsWith(value.toUpperCase())`

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,166 @@
...
+    const listElmt = document.createElement('div');
+    listElmt.innerHTML = '<strong>' +
+                         arrElt.substr(0, val.length) +
+                         '</strong>';
+    listElmt.innerHTML += arrElt.substr(val.length);
+    const cpyArrElt = arrElt;
+    listElmt.addEventListener('click', () => {
+      SEARCH_INPUT.value = cpyArrElt;
+      keepTrackOfSuggestions(cpyArrElt);
+      closeAllElmntExcept();
```
fabielzuniga, line 110: I would suggest adding two methods. It reads a little strange the see the 'except' suffix and then passing no arguments. One method could be closeAllElements() and the other closeAllElementsExcept(element)
Thoughts?

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,166 @@
...
+    if (arrElt.substr(0, val.length).toUpperCase() === val.toUpperCase()) {
+      relevantSuggestions.push(arrElt);
+    }
+  }
+  return relevantSuggestions;
+}
+
+/**
+ * Takes relevant suggestions and displays them in DOM
+ * @param {Array<string, CollegeId>} relevantSuggestions
```
fabielzuniga, line 95: `@param {Array<string>} relevantSuggestions` ?

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -0,0 +1,166 @@
...
+      closeAllElmntExcept();
+    });
+    autocompleteList.appendChild(listElmt);
+  }
+}
+
+/**
+ * Event occurrance when input is provided to autocomplete field
+ */
+function getsAllProposedSuggestions() {
```
fabielzuniga, line 119: get instead of gets?

capstone/src/main/webapp/js/autocomplete.js
mmurray22: thanks! done!

capstone/src/main/webapp/js/autocomplete.js
mmurray22: ah thanks! done!

capstone/src/main/webapp/js/autocomplete.js
mmurray22: ah thanks! done!

capstone/src/main/webapp/js/autocomplete.js
mmurray22: ah cool! did not know about that :) Thanks!

capstone/src/main/webapp/js/autocomplete.js
mmurray22: that would make it smoother :) can do!

capstone/src/main/webapp/js/autocomplete.js
mmurray22: true!

PR #52, opened 2020-08-04
CollegeDataApi.java class interface

biancamacias:
This is just an interface for the CollegeDataApi.java class, which is meant to be the only class that interacts with the CollegeScorecard API. I have added some abstract methods and descriptions. 

This interface has only one method, ```getCollegesById```, which is meant to send the REST request into the backend and use the converter to convert from ```ScorecardData``` class to ```CollegeData``` class.


capstone/src/main/java/com/google/univiz/scorecard/CollegeData.java
```diff
@@ -0,0 +1,23 @@
+package com.google.univiz.scorecard;
+
+/**
+ * CollegeData is an interface that contains methods and functions which communicate with the
+ * CollegeScorecardAPI to obtain college data from the URL in JSON format.
+ */
+public interface CollegeData {
+
+  /** TODO(biancamacias): decide what getSuggestions does */
```
mmurray22, line 10: we can talk about this if you want :)

capstone/src/main/java/com/google/univiz/scorecard/CollegeData.java
```diff
@@ -0,0 +1,23 @@
...
+   * getCollegesById takes ids and field names of colleges selected and return a list of
+   * ScorecardData to be used by the servlets.
+   */
+  public List<ScorecardData> getCollegesById(List<String> ids, List<String> fieldNames) {}
+
+  /**
+   * scorecardApiCall takes ids and field names of colleges selected to call the CollegeScorecardAPI
+   * and return the data as JSON.
+   */
+  public String scorecardApiCall(List<String> ids, List<String> fieldNames) {}
```
mmurray22, line 23: is this supposed to return a JSON object?

capstone/src/main/java/com/google/univiz/scorecard/CollegeData.java
```diff
@@ -0,0 +1,23 @@
...
+public interface CollegeData {
+
+  /** TODO(biancamacias): decide what getSuggestions does */
+  public List<ScorecardData> getSuggestions(String suggestion) {}
+
+  /**
+   * getCollegesById takes ids and field names of colleges selected and return a list of
+   * ScorecardData to be used by the servlets.
+   */
+  public List<ScorecardData> getCollegesById(List<String> ids, List<String> fieldNames) {}
```
mmurray22, line 17: one suggestion: you could make a custom class to hold both IDs and Names

capstone/src/main/java/com/google/univiz/scorecard/CollegeData.java
```diff
@@ -0,0 +1,23 @@
+package com.google.univiz.scorecard;
+
+/**
```
mmurray22, line 3: does this need any JavaDoc? not sure

capstone/src/main/java/com/google/univiz/scorecard/CollegeData.java
biancamacias: yeah a string that holds the JSON response? but i havent done the implementation, i might have to use the ```JSONobject``` type.

capstone/src/main/java/com/google/univiz/scorecard/CollegeData.java
```diff
@@ -0,0 +1,31 @@
...
+
+  /**
+   * Returns a list of ScorecardData to be used by the servlets.
+   *
+   * @param ids A list of college IDs that were selected by the user.
+   * @param fieldNames A list of fieldNames of data requested that will be used to call the
+   *     CollegeScorecard API?
+   * @return A list of ScorecardData types to be used in the back-end.
+   */
+  public List<ScorecardData> getCollegesById(List<String> ids, List<String> fieldNames) {}
```
ihsan314, line 21: In [Kaveh's layout](https://docs.google.com/document/d/1G2QrgDMhVCPX6cbNOQNKvexDImB5Duw1dr50cDGx70U/edit?disco=AAAAKDmwaXI) he suggested a `@Nullable String sortByField` parameter for this function, might be useful if we want to deal with data based on how it's sorted under the specified parameter. I don't know if it's necessary for the function of this though, but it's probably worth consideration.

capstone/src/main/java/com/google/univiz/scorecard/CollegeData.java
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * CollegeData is an interface that contains methods and functions which communicate with the
+ * CollegeScorecardAPI to obtain college data from the URL in JSON format and parse them to become
+ * ScorecardData java objects.
+ */
+public interface CollegeData {
+
+  /** TODO(biancamacias): decide what getSuggestions does */
+  public List<ScorecardData> getSuggestions(String suggestion) {}
```
ihsan314, line 11: If this is an interface, since you're not going to provide a default implementation here, then I believe you can do this instead: 
```suggestion
  public List<ScorecardData> getSuggestions(String suggestion);
```
Likewise for the remaining functions.

capstone/src/main/java/com/google/univiz/scorecard/CollegeData.java
ccondit: I would leave it in. it's good to know what this interface intends to do...

capstone/src/main/java/com/google/univiz/scorecard/CollegeData.java
```diff
@@ -0,0 +1,33 @@
+package com.google.univiz.scorecard;
+
+import java.util.List;
+
+/**
+ * CollegeData is an interface that contains methods and functions which communicate with the
+ * CollegeScorecardAPI to obtain college data from the URL in JSON format and parse them to become
```
ccondit, line 7: nit - i would leave out implementation details here. the point of the interface is that it should be agnostic to collegescorecardapi.

capstone/src/main/java/com/google/univiz/scorecard/CollegeData.java
```diff
@@ -0,0 +1,33 @@
...
+ * ScorecardData java objects.
+ */
+public interface CollegeData {
+
+  /** TODO(biancamacias): decide what getSuggestions does */
+  public List<ScorecardData> getSuggestions(String suggestion);
```
ccondit, line 13: nit - interface methods are `public` by default. can you also remove the one below?
```suggestion
  List<ScorecardData> getSuggestions(String suggestion);
```

ccondit, line 13:
```suggestion
  public List<ScorecardData> getSuggestions(String query);
```
nit - be more specific in the method argument name.

ccondit, line 13: I think we'll want a dedicated type for autocomplete and not ScoreCard data: or- can you say more about what you intend to return here?

capstone/src/main/java/com/google/univiz/scorecard/CollegeData.java
```diff
@@ -0,0 +1,33 @@
...
+
+  /**
+   * Returns a list of ScorecardData to be used by the servlets.
+   *
+   * @param ids A list of college IDs that were selected by the user.
+   * @param fieldNames A list of fieldNames of data requested that will be used to call the
+   *     CollegeScorecard API?
+   * @return A list of ScorecardData types to be used in the back-end.
+   */
+  public List<ScorecardData> getCollegesById(List<String> ids, List<String> fieldNames);
```
ccondit, line 23: Is this going to be a String for the college id? or a CollegeId class?

Also - why is the fieldnames argument necessary. Won't the implementation handle this?

capstone/src/main/java/com/google/univiz/scorecard/CollegeData.java
```diff
@@ -0,0 +1,33 @@
...
+
+  /**
+   * Returns CollegeScorecard API call as JSON.
+   *
+   * @param ids A list of college IDs that were selected by the user.
+   * @param fieldNames A list of fieldNames of data that will be used in the URL that calls the API.
+   * @return A string that holds the JSON response.
+   */
+  public String scorecardApiCall(List<String> ids, List<String> fieldNames);
```
ccondit, line 32: i would remove this method since it's implementation specific.

capstone/src/main/java/com/google/univiz/scorecard/CollegeData.java
kaveh096, line 23: Both of these arguments could be a `Set<>` which gives us better guarantees.
```suggestion
  public List<ScorecardData> getCollegesById(Set<String> ids, Set<String> fieldNames);
```

capstone/src/main/java/com/google/univiz/scorecard/CollegeData.java
kaveh096: [#leaky_abstraction](https://en.wikipedia.org/wiki/Leaky_abstraction)

capstone/src/main/java/com/google/univiz/scorecard/CollegeData.java
biancamacias: i was going off of the methods kaveh laid out in the data flow doc, im not entirely sure. we can talk about this offline.

capstone/src/main/java/com/google/univiz/CollegeDataApi.java
```diff
@@ -0,0 +1,9 @@
+package com.google.univiz;
+
+import java.util.List;
+import java.util.Set;
+
+public interface CollegeDataApi {
+
+  List<CollegeData> getCollegesById(Set<String> ids, Set<String> names);
```
ccondit, line 8: Here a String is used for the college id, but in CollegeData, it's an int. Which should we use? Or should there be a dedicated `CollegeId class`?

ccondit, line 8: Why are `names` required? How will they be used?

capstone/src/main/java/com/google/univiz/CollegeData.java
```diff
@@ -0,0 +1,50 @@
+package com.google.univiz;
+
+import com.google.auto.value.AutoValue;
+
+/**
+ * CollegeData is a class representing a single university/college/institution. These parameters use
+ * data parsed from the CollegeScorecard API. This data has been converted from the ScorecardData
```
ccondit, line 7: i wouldn't mention scorecard here since it's meant to be agnostic of what implementation is providing the data. These sort of comments would be great in the implementation class that implements this interface.

capstone/src/main/java/com/google/univiz/CollegeData.java
```diff
@@ -0,0 +1,50 @@
...
+  /**
+   * Returns a boolean that tells whether the institution is the main campus or not, true or false.
+   */
+  abstract boolean isMainCampus();
+
+  /**
+   * Returns a string that describes the institution's degree of urbanzation, or how populated the
+   * surrounding areas are around the institution, from large city to rural.
+   */
+  abstract String urbanizationDegree();
```
ccondit, line 28: This would be better modeled as an Enum with the potential values of the urbanization degree.

capstone/src/main/java/com/google/univiz/CollegeData.java
```diff
@@ -0,0 +1,50 @@
...
+  abstract float latitude();
+
+  abstract float longitude();
+
+  /**
+   * Returns a string that represents the institution's carnegie size degree, or the highest degree
+   * certificate available at that institution, from not applicable to exclusively
+   * graduate/professional.
+   */
+  abstract String carnegieSizeDegree();
```
ccondit, line 39: Same here, does it make sense to model this as an Enum? I think it will make the parsing easier, too...

capstone/src/main/java/com/google/univiz/CollegeData.java
biancamacias: anything that has some range of values ill make sure to make an Enum.

capstone/src/main/java/com/google/univiz/CollegeDataApi.java
kaveh096: Back when i was thinking about this method, i assumed it would get the list of field names that it should return. But i believe we later decided that it should return all fields and later resource classes would down-sample it to the desired autovalue. So the second argument was the desired field names. For example MapResouce class would call this method and only ask for lat, lng, and name fields to be returned. You can remove it for now and if we wanted to take this route (to improve performance) we can add it back later.

capstone/src/main/java/com/google/univiz/CollegeData.java
```diff
@@ -0,0 +1,86 @@
...
+  abstract int numOfUndergrads();
+
+  abstract int avgCost();
+
+  abstract float ratioOfMen();
+
+  abstract float ratioOfWomen();
+
+  static Builder builder() {
+    return new AutoValue_CollegeData.Builder().setUrbanizationDegree(0);
```
ihsan314, line 50: This line may need to change to 
```suggestion
    return new AutoValue_CollegeData.Builder().setUrbanizationDegree(UrbanizationDegree.NOT_AVAILABLE);
```
or something similar since likely now the integer 0 doesn't directly translate to an `UrbanizationDegree` anymore. But even if this code is still valid, the above suggestion makes use of the name of the type, and may be more readable.

capstone/src/main/java/com/google/univiz/CarnegieSizeDegree.java
```diff
@@ -0,0 +1,31 @@
+package com.google.univiz;
+
+/**
+ * CarnegieSizeDegree is a class that holds the enum for the range of values for a school's carnegie
+ * size degree. The index of each of these values is the carnegie size degree provided by
+ * ScorecardData.
+ */
+public class CarnegieSizeDegree {
```
ihsan314, line 8: Actually nvm I think you no longer need this, since `enum CarnegieSizeDegree` is syntactic sugar for `class CarnegieSizeDegree extends Enum`.

capstone/src/main/java/com/google/univiz/UrbanizationDegree.java
```diff
@@ -0,0 +1,21 @@
+package com.google.univiz;
+
+/**
+ * UrbanizationDegree is a class that holds the enum for the range of values for a school's
+ * urbanization degree. The index of each of these values is the urbanization degree provided by
+ * ScorecardData.
+ */
+public class UrbanizationDegree {
```
ihsan314, line 8: Ditto.

ihsan314:
Also you're missing a constructor and a member variable. See the `Planet` class [here](https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html) for more on creating an Enum.

capstone/src/main/java/com/google/univiz/CollegeData.java
```diff
@@ -0,0 +1,76 @@
+package com.google.univiz;
+
+import com.google.auto.value.AutoValue;
+
+/** CollegeData is a class representing a single university/college/institution. */
+@AutoValue
+abstract class CollegeData {
+
+  abstract int id();
```
ccondit, line 9: I suspect you'll want this to be `CollegeId`, which is introduced [here](https://github.com/googleinterns/univiz/blob/1d1061804b2bc9003da2aae7e9b850c6fe3c3823/capstone/src/main/java/com/google/univiz/api/CollegeId.java).
```suggestion
  abstract CollegeId id();
```

capstone/src/main/java/com/google/univiz/CollegeDataApi.java
```diff
@@ -0,0 +1,13 @@
+package com.google.univiz;
+
+import java.util.List;
+import java.util.Set;
+
+/**
+ * The CollegeDataApi interface is responsible for all REST requests and JSON parsing to the
+ * CollegeData class. CollegeDataApi is the only class that will communicate with an external API.
```
ccondit, line 8:
```suggestion
 * CollegeData class.
```

I wouldn't make claims that might change in the future :)

capstone/src/main/java/com/google/univiz/CollegeDataApi.java
```diff
@@ -0,0 +1,13 @@
...
+ */
+public interface CollegeDataApi {
+
+  List<CollegeData> getCollegesById(Set<String> ids);
```
ccondit, line 12:
```suggestion
  List<CollegeData> getCollegesById(Set<CollegeId> ids);
```

capstone/src/main/java/com/google/univiz/CarnegieSizeDegree.java
```diff
@@ -0,0 +1,34 @@
...
+  SIZE11("Four-year college, small and residential"),
+  SIZE12("Four-year college, medium and mostly nonresidential"),
+  SIZE13("Four-year college, medium and mostly residential"),
+  SIZE14("Four-year college, medium and residential"),
+  SIZE15("Four-year college, large and mostly nonresidential"),
+  SIZE16("Four-year college, large and mostly residential"),
+  SIZE17("Four-year college, large and residential"),
+  SIZE18("Exclusively graduate and professial institution");
+
+  private final String degree;
```
ccondit, line 29: Is there going to be a way for clients to get this field? Consider adding a `getter` method.

capstone/src/main/java/com/google/univiz/CarnegieSizeDegree.java
```diff
@@ -0,0 +1,34 @@
+package com.google.univiz;
+
+/**
+ * CarnegieSizeDegree is a class that holds the enum for the range of values for a school's carnegie
+ * size degree. The index of each of these values is the carnegie size degree provided by
+ * ScorecardData.
```
ccondit, line 6: Since this is in UniViz, let's leave scorecard details out...

```suggestion
 * size degree.
```

biancamacias:
> I concur with what Ihsan is saying about enums-looking good overall! :) I also have a general question about the focus of the PR: based on the PR description, this is supposed to be just for an interface for the class which will actively interact with the REST API. However, it seems to focus more on the CollegeData class now. Splitting this PR into 1 PR for the CollegeDataAPI interface and 1 PR for the CollegeData type/related enums would be helpful because otherwise it seems that the CollegeDataAPI gets lost in all the CollegeData changes

ahhhh, i see what youre saying. keep this class as strictly the interface, and move the CollegeData into its own PR. i will do that. makes more sense. i guess i added it here because the interface uses `CollegeData`

capstone/src/main/java/com/google/univiz/CollegeDataApi.java
```diff
@@ -0,0 +1,13 @@
...
+import java.util.List;
+import java.util.Set;
+
+/**
+ * The CollegeDataApi interface is responsible for all REST requests and JSON parsing to the
+ * CollegeData class.
+ */
+public interface CollegeDataApi {
+
+  List<CollegeData> getCollegesById(Set<CollegeId> ids);
```
ihsan314, line 12: Did we already merge CollegeId into master? If not, merging the repo as is will break future builds unnecessarily.

kaveh096:
> > I concur with what Ihsan is saying about enums-looking good overall! :) I also have a general question about the focus of the PR: based on the PR description, this is supposed to be just for an interface for the class which will actively interact with the REST API. However, it seems to focus more on the CollegeData class now. Splitting this PR into 1 PR for the CollegeDataAPI interface and 1 PR for the CollegeData type/related enums would be helpful because otherwise it seems that the CollegeDataAPI gets lost in all the CollegeData changes
> 
> ahhhh, i see what youre saying. keep this class as strictly the interface, and move the CollegeData into its own PR. i will do that. makes more sense. i guess i added it here because the interface uses `CollegeData`

If you want to split this PR into two, then the data structure should come first then the api PR. Otherwise in the API PR you don't have access to the return type (CollegeData) definition.

biancamacias:
> > > I concur with what Ihsan is saying about enums-looking good overall! :) I also have a general question about the focus of the PR: based on the PR description, this is supposed to be just for an interface for the class which will actively interact with the REST API. However, it seems to focus more on the CollegeData class now. Splitting this PR into 1 PR for the CollegeDataAPI interface and 1 PR for the CollegeData type/related enums would be helpful because otherwise it seems that the CollegeDataAPI gets lost in all the CollegeData changes
> > 
> > 
> > ahhhh, i see what youre saying. keep this class as strictly the interface, and move the CollegeData into its own PR. i will do that. makes more sense. i guess i added it here because the interface uses `CollegeData`
> 
> If you want to split this PR into two, then the data structure should come first then the api PR. Otherwise in the API PR you don't have access to the return type (CollegeData) definition.

i see. ill rebase this PR then. thank you.

PR #55, opened 2020-08-05
simple map display

biancamacias:
This PR creates and displays a map onto ```dashboard.html```. This helped visualize where the map would be on the webpage, and it helped test the API key. The center location is hardcoded, and the ```script src=``` reference is subject to change if we decide to combine all JS files. the ```onload``` in the ```<body>``` is also subject to change to include other functions other than ```createMap()```. 

the lint problems at the moment are that ```createMap()``` is not used and that the ```const map``` is not used.

This is what the dashboard looks like:

image

### DIFF
--- a/capstone/src/main/webapp/js/maps.js
+++ b/capstone/src/main/webapp/js/maps.js
@@ -1,10 +1,9 @@
 /**
- * createMap() just displays the map onto the dashboard.
- * All parameters for positioning are hardcoded.
+ * Displays the map onto the dashboard with hardcoded values.
+ * @returns {void}
  */
-
 function createMap() {
   const map = new google.maps.Map(
-    document.getElementById('map'),
-    {center: {lat: 34.052, lng: -118.245}, zoom: 10});
+      document.getElementById('map'),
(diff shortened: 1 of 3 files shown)


capstone/src/main/webapp/js/maps.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * Displays the map onto the dashboard with hardcoded values.
+ * @return {void}
```
ihsan314, line 3: I believe the `@return` tag is only needed if the function actually returns a parameter. If you want you can keep it though.

capstone/src/main/webapp/js/maps.js
```diff
@@ -0,0 +1,9 @@
...
+ */
+function createMap() {
+  const map = new google.maps.Map(
+      document.getElementById('map'),
+      {center: {lat: 34.052, lng: -118.245}, zoom: 10});
```
ihsan314, line 8: Not important now but later when you are fetching data you could possibly center the map on the average latitude-longitude pair of the user's selected colleges.

PR #57, opened 2020-08-06
Completes Feature Interfaces

mmurray22:
Adds content to SearchResource and VisResource. Adds relevant object files. 

Note: This PR depends on the CollegeData data type to compile (which I believe will be worked on in a future PR). Currently I have dummy CollegeData/SuggestionData classes to allow for compilation


capstone/src/main/java/com/google/univiz/api/SearchResource.java
```diff
@@ -1,7 +1,15 @@
...
  * The SearchResoure interface is responsible for all data management/processing specific for the
  * search feature of the UniViz webapp.
  */
-public interface SearchResource {}
+public interface SearchResource {
+  /**
+   * Take a list of suggestion data and returns a list of Suggestions objects to the caller. This
+   * object contains all data relevant to the Search feature.
+   */
+  List<SearchData> getSearchSuggestions(List<SuggestionData> collegeNames);
```
ccondit, line 14: IIUC `SearchResource.getSearchSuggestions` will be invoked from a `SearchServlet`.  How will the servlet produce `SuggestionData`? Should this method take a String argument, instead?

mmurray22:
so should I change this PR to be just about SearchResource and then another PR just about VisResource? @ccondit

capstone/src/main/java/com/google/univiz/api/SearchResource.java
mmurray22: OMG yes-I confused myself here too for this PR and the other one chained with this 

that makes a lot more sense-will update!

capstone/src/main/java/com/google/univiz/api/SearchResource.java
```diff
@@ -1,7 +1,15 @@
...
  * The SearchResoure interface is responsible for all data management/processing specific for the
  * search feature of the UniViz webapp.
  */
-public interface SearchResource {}
+public interface SearchResource {
+  /**
+   * Take a list of suggestion data and returns a list of Suggestions objects to the caller. This
+   * object contains all data relevant to the Search feature.
+   */
+  List<SearchData> getSearchSuggestions(List<String> collegeNames);
```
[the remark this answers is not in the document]
kaveh096: Same question. I also thought this would be a single string that the user typed in the search box.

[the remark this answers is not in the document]
kaveh096: Same question. I also thought this would be a single string that the user typed in the search box.

capstone/src/main/java/com/google/univiz/api/SearchResource.java
```diff
@@ -1,7 +1,15 @@
...
+
 /**
  * The SearchResoure interface is responsible for all data management/processing specific for the
  * search feature of the UniViz webapp.
  */
-public interface SearchResource {}
+public interface SearchResource {
+  /**
+   * Take a list of suggestion data and returns a list of Suggestions objects to the caller. This
+   * object contains all data relevant to the Search feature.
```
kaveh096, line 12: Is this the method that we'll use for search auto-complete feature? If yes, here is my suggestion for the javadoc.

```suggestion
   * Given a potentially incomplete search query, returns a list of colleges that would match the query.
   *
   * @param query The search query as the user typed in. It should be a non-empty and a non-null string.
   * @return An ordered list of possible matching colleges. The list is ordered such that the best match is at the
   *   beginning of the list. The list may be empty which indicates there are no possible matches. The list is also
   *   not an exhaustive set of matches. It only includes few top results.
```

Please fix formatting as needed.

capstone/src/main/java/com/google/univiz/api/CollegeData.java
```diff
@@ -0,0 +1,4 @@
+package com.google.univiz.api;
+
+/** THIS IS A DUMMY CLASS TO ALLOW COMPILATION: REMOVE BEFORE MERGING */
+public class CollegeData {}
```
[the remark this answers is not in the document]
mmurray22: deleted!

capstone/src/main/java/com/google/univiz/api/SearchResource.java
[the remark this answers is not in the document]
mmurray22: updated 👍

capstone/src/main/java/com/google/univiz/api/SearchResource.java
```diff
@@ -1,7 +1,22 @@
...
  * search feature of the UniViz webapp.
  */
-public interface SearchResource {}
+public interface SearchResource {
+  /**
+   * Given a potentially incomplete search query, returns a list of colleges that would match the
+   * query.
+   *
+   * @param query The search query as the user typed in. It should be a non-empty and a non-null
+   *     string.
```
ccondit, line 15: A few nits on this param doc:
* "as the user typed in" - in the scope of this project, this will probably be the case, but we could imagine other uses for this interface (ie a batch job trying to match text).
* what would the consequences be of invoking this method with an empty or null string? we have a couple of options here:
  - document what happens in this situations
  - be defensive so the client doesn't need to worry about this (ie empty input yields empty results).

capstone/src/main/java/com/google/univiz/api/SearchResource.java
```diff
@@ -1,7 +1,22 @@
...
+   * @return An ordered list of possible matching colleges. The list is ordered such that the best
```
ccondit, line 16:
```suggestion
   * @return A list of possible matching colleges. The list is ordered such that the best
```

Lists are inherently ordered :)

capstone/src/main/java/com/google/univiz/api/SearchResource.java
```diff
@@ -1,7 +1,22 @@
...
+   *     match is at the beginning of the list. The list may be empty which indicates there are no
+   *     possible matches. The list is also not an exhaustive set of matches. It only includes few
+   *     top results.
```
ccondit, line 19: > It only includes few top results

This is a great point - do you think this should be something that a client can configure? If so, consider adding it as an argument to this method.

mmurray22: Probably! Ok, sounds good :) is that something I should handle in the front-end/servlet or backend?

capstone/src/main/java/com/google/univiz/api/SearchResource.java
mmurray22: I would probably go with the defensive strategy

capstone/src/main/java/com/google/univiz/api/SearchResource.java
```diff
@@ -1,7 +1,22 @@
...
+   * query.
+   *
+   * @param query The search query as the user provided. It should be a non-empty and a non-null
+   *     string.
+   * @return A list of possible matching colleges. The list is ordered such that the best match is
+   *     at the beginning of the list. The list may be empty which indicates there are no possible
+   *     matches. The list is also not an exhaustive set of matches. It only includes few top
+   *     results. If the string is invalid (empty, null, etc.), an empty list will be returned
+   */
+  List<SearchData> getSearchSuggestions(String collegeNames);
```
kaveh096, line 21: From the javadoc, it looks like you're going with the parameter name `query`.

```suggestion
  List<SearchData> getSearchSuggestions(String query);
```

ccondit: +1

capstone/src/main/java/com/google/univiz/api/SearchResource.java
```diff
@@ -1,7 +1,22 @@
...
+  /**
+   * Given a potentially incomplete search query, returns a list of colleges that would match the
+   * query.
+   *
+   * @param query The search query as the user provided. It should be a non-empty and a non-null
+   *     string.
+   * @return A list of possible matching colleges. The list is ordered such that the best match is
+   *     at the beginning of the list. The list may be empty which indicates there are no possible
+   *     matches. The list is also not an exhaustive set of matches. It only includes few top
+   *     results. If the string is invalid (empty, null, etc.), an empty list will be returned
```
ccondit, line 19: if which string is invalid?

mmurray22: done!

mmurray22:
Hi Chris, what did the +1 mean?

ccondit:
> Hi Chris, what did the +1 mean?

I was attempting to +1 this:

https://github.com/googleinterns/univiz/pull/57#discussion_r468993053

mmurray22:
Ah ok-updated @ccondit!

PR #58, opened 2020-08-06
Adds search interface implementation

mmurray22:
Implementation for the search interface. The search interface is part of the autocomplete pipeline, and this is the stage which obtains the data from the REST Api Library and sends it to the front-end servlet. 

Files Included:
-SearchResourceImpl.java
-SearchResourceImplTest.java
-SearchData.java

Updates:
Now with Guice :)


capstone/src/main/java/com/google/univiz/api/SearchData.java
```diff
@@ -1,4 +1,12 @@
 package com.google.univiz.api;
 
 /** SearchData represents the college data specific to the Search feature */
-public class SearchData {}
+public class SearchData {
+  String collegeName;
+  int collegeId;
```
ihsan314, line 6:
```suggestion
  private String collegeName;
  private int collegeId;
```

capstone/src/main/java/com/google/univiz/api/SearchData.java
```diff
@@ -1,4 +1,12 @@
 package com.google.univiz.api;
 
 /** SearchData represents the college data specific to the Search feature */
-public class SearchData {}
```
biancamacias, line 4: i think we want to use ```AutoValue``` for the ```SearchData``` and ```SuggestionData``` class. look at #37 for reference on how to use it, and if you need some help, put it in the chat. also #52 has another ```AutoValue``` example.

capstone/src/main/java/com/google/univiz/api/SuggestionData.java
```diff
@@ -1,4 +1,20 @@
 package com.google.univiz.api;
 
 /** THIS IS A DUMMY CLASS TO ALLOW COMPILATION: REMOVE BEFORE MERGING */
-public class SuggestionData {}
```
biancamacias, line 4: ditto about ```AutoValue``` here

capstone/src/main/java/com/google/univiz/api/SearchResourceImpl.java
```diff
@@ -0,0 +1,17 @@
+package com.google.univiz.api;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/** Implementation of SearchResource Interface */
+public class SearchResourceImpl extends SearchResource {
```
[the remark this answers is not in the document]
biancamacias: i think this is because chris mentioned that adding ```Impl``` is common for a class that implements an interface.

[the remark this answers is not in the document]
mmurray22: ^^ yep this is the standard naming convention. Thanks for catching the error Ihsan!

capstone/src/main/java/com/google/univiz/api/SuggestionData.java
```diff
@@ -1,4 +1,20 @@
 package com.google.univiz.api;
 
 /** THIS IS A DUMMY CLASS TO ALLOW COMPILATION: REMOVE BEFORE MERGING */
```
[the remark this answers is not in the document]
mmurray22: most likely yes

capstone/src/main/java/com/google/univiz/api/SuggestionData.java
mmurray22: ah ok-should I go through the work of doing autovalue stuff here since this is purely for compilation?

capstone/src/main/java/com/google/univiz/api/SearchData.java
mmurray22: ah will do!

capstone/src/main/java/com/google/univiz/api/SuggestionData.java
biancamacias: no, but if you're going to have this class in another PR, just remember to use ```AutoValue```

capstone/src/main/java/com/google/univiz/api/SearchData.java
```diff
@@ -1,4 +1,18 @@
...
+@AutoValue
+public abstract class SearchData {
+  private String collegeName;
+  private int collegeId;
+
+  public static SearchData create(String collegeName, int collegeId) {
+    return new AutoValue_SearchData(collegeName, collegeId);
+  }
```
biancamacias, line 13:
```suggestion

```
you don't need this, just the accessor methods is fine.

capstone/src/main/java/com/google/univiz/api/fix.sh
```diff
@@ -0,0 +1,13 @@
+
```
ihsan314, line 1: One way to prevent accidentally committing this file is making a `bin` directory inside your home directory. Then you could add this script there and add the bin directory to your `PATH`, which would make `fix.sh` accessible throughout.

capstone/src/main/java/com/google/univiz/api/SearchData.java
mmurray22: I don't think I need the private variables at the beginning of the class, but don't I still need a create the SearchData obj?

capstone/src/main/java/com/google/univiz/api/fix.sh
mmurray22: yep that's a great point because I always end up accidentally committing this :((

capstone/src/main/java/com/google/univiz/api/SearchData.java
```diff
@@ -1,4 +1,15 @@
...
 
+import com.google.auto.value.AutoValue;
+
 /** SearchData represents the college data specific to the Search feature */
-public class SearchData {}
+@AutoValue
+public abstract class SearchData {
+  public static SearchData create(String collegeName, int collegeId) {
+    return new AutoValue_SearchData(collegeName, collegeId);
+  }
```
biancamacias, line 10:
```suggestion

```

mmurray22: Hi, as per my previous comment, I'm still wondering why I need to delete this? I don't doubt that this is a correct way to do it, but I don't understand why it needs to be done :)

ccondit: I think you want to leave this in, since you'll create these instances from `SearchResourceImpl`

capstone/src/main/java/com/google/univiz/api/SearchResource.java
```diff
@@ -11,5 +11,5 @@
    * Take a list of suggestion data and returns a list of Suggestions objects to the caller. This
    * object contains all data relevant to the Search feature.
    */
-  List<SearchData> getSearchSuggestions(List<SuggestionData> collegeNames);
+  List<SearchData> getSearchSuggestions(List<String> collegeNames);
```
ccondit, line 14: How should the client supply the `List<String>` in this method signature? The client just has free text from the autocomplete box. How (or should?) will the client convert that into a list?

capstone/src/main/java/com/google/univiz/api/SuggestionData.java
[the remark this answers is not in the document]
ccondit: Seems like this isn't the case anymore? I'm also not clear on how this class differs from `SearchData`

capstone/src/main/java/com/google/univiz/api/SearchResourceImpl.java
```diff
@@ -0,0 +1,19 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+/** Implementation of SearchResource Interface */
+public class SearchResourceImpl implements SearchResource {
+
+  public List<SearchData> getSearchSuggestions(List<String> collegeNames) {
+    List<SearchData> searchDataList = new ArrayList<>();
+    List<SuggestionData> collegeSuggestions = getSuggestions(collegeNames);
```
ccondit, line 11: What is `getSuggestions`?

capstone/src/main/java/com/google/univiz/api/SearchResourceImpl.java
```diff
@@ -0,0 +1,19 @@
+package com.google.univiz.api;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/** Implementation of SearchResource Interface */
+public class SearchResourceImpl implements SearchResource {
```
ccondit, line 7: Can you add a test for this class?

capstone/src/main/java/com/google/univiz/api/SearchResource.java
mmurray22: client shouldn't be converting it into a list :) just need a single String!

capstone/src/main/java/com/google/univiz/api/SearchResourceImpl.java
mmurray22: this would be a function obtained from a function in SuggestionAPI, which communicates with the REST API. Should I add a comment here?

capstone/src/main/java/com/google/univiz/api/SearchResourceImpl.java
mmurray22: yes indeed!

capstone/src/main/java/com/google/univiz/api/SearchData.java
```diff
@@ -1,4 +1,15 @@
...
+
+  public abstract String getCollegeName();
+
+  public abstract int getCollegeId();
```
kaveh096, line 14: Should this return an instance of CollegeId?

capstone/src/main/java/com/google/univiz/api/SearchData.java
```diff
@@ -1,4 +1,15 @@
 package com.google.univiz.api;
 
+import com.google.auto.value.AutoValue;
+
 /** SearchData represents the college data specific to the Search feature */
-public class SearchData {}
+@AutoValue
+public abstract class SearchData {
+  public static SearchData create(String collegeName, int collegeId) {
```
kaveh096, line 8: I don't think this static method needs to be public. Only api classes would ever need to create an instance of this class.
```suggestion
  static SearchData create(String collegeName, int collegeId) {
```

capstone/src/main/java/com/google/univiz/api/SearchResource.java
```diff
@@ -11,5 +11,5 @@
    * Take a list of suggestion data and returns a list of Suggestions objects to the caller. This
    * object contains all data relevant to the Search feature.
    */
-  List<SearchData> getSearchSuggestions(List<SuggestionData> collegeNames);
+  List<SearchData> getSearchSuggestions(String collegeNames);
```
kaveh096, line 14: consider a better more descriptive name for this parameter. Like collegeName, partialCollegeName, collegeNameQuery, searchQuery, etc.

capstone/src/main/java/com/google/univiz/api/SearchResource.java
```diff
@@ -11,5 +11,5 @@
    * Take a list of suggestion data and returns a list of Suggestions objects to the caller. This
    * object contains all data relevant to the Search feature.
```
kaveh096, line 12: Now that you've fixed the parameter type, the javadoc also needs to be updated.

capstone/src/main/java/com/google/univiz/api/SearchResourceImpl.java
```diff
@@ -0,0 +1,19 @@
+package com.google.univiz.api;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/** Implementation of SearchResource Interface */
```
kaveh096, line 6: This javadoc currently does not add any value. it's documenting something that is pretty obvious. Either remove it or add more details.

capstone/src/main/java/com/google/univiz/api/SearchResourceImpl.java
kaveh096, line 7: I think this class should remain package-private.

```suggestion
class SearchResourceImpl implements SearchResource {
```

capstone/src/main/java/com/google/univiz/api/SearchResourceImpl.java
```diff
@@ -0,0 +1,19 @@
...
+public class SearchResourceImpl implements SearchResource {
+
+  public List<SearchData> getSearchSuggestions(List<String> collegeNames) {
```
kaveh096, line 9: Method implementation should match interface
```suggestion
  @Override
  public List<SearchData> getSearchSuggestions(String collegeNameQuery) {
```

capstone/src/main/java/com/google/univiz/api/SearchResourceImpl.java
kaveh096: So for you to be able to use this method, first you have to add the SuggestionAPI as a field in this class. But that API is not yet in master, so first you have to wait for the corresponding PR to be merged. All you can do right now is to add a TODO.

capstone/src/main/java/com/google/univiz/api/SearchData.java
mmurray22: no-in this case, it would be a getter function for an inner value in the SearchData obj. Should I change the naming scheme/the object?

capstone/src/main/java/com/google/univiz/api/SearchResource.java
mmurray22: updated!

capstone/src/main/java/com/google/univiz/api/SearchResource.java
mmurray22: updated!

capstone/src/main/java/com/google/univiz/api/SearchResourceImpl.java
mmurray22: removed!

biancamacias:
Looks like the build errors are from the missing `SuggestionData` class, so I think this will be resolved once you merge. LGTM.

mmurray22:
@ccondit and @kaveh096 Here is the progress I have so far on this branch. I believe this is the branch that could benefit from Guice injection, so I will be working on Guice next

capstone/src/main/java/com/google/univiz/api/SearchResourceImpl.java
```diff
@@ -0,0 +1,25 @@
...
+import javax.inject.Inject;
+import java.util.ArrayList;
+import java.util.List;
+
+class SearchResourceImpl implements SearchResource {
+  private List<SuggestionData> collegeSuggestions;
+
+  @Inject
+  public void Restaurant(collegeSuggestions) {
+    this.collegeSuggestions = collegeSuggestions;
```
ihsan314, line 12:
```suggestion
  public SearchResourceImpl(List<SuggestionData> collegeSuggestions) {
      this.collegeSuggestions = new ArrayList<>();
      this.collegeSuggestions.addAll(collegeSuggestions);
```
Few things here:

- This looks like a constructor, and if that assumption is correct, then the name and return type were off.
- The assignment here is a shallow assignment, which is fine sometimes, but it also means anyone/any class with the reference to `collegeSuggestions` will be able to manipulate the instance's copy of `collegeSuggestions` even after the instance has been created. Given that `collegeSuggestions` is a private member, it's probably a good idea to make sure it only gets manipulated by `SearchResourceImpl`'s methods.
- To get Guice to know how to inject `collegeSuggestions` remember to write a subclass of `AbstractModule`.

mmurray22:
Not done with the test yet-if you have any comments don't comment on the tests yet! Thanks!

capstone/src/main/java/com/google/univiz/api/SearchResourceImpl.java
```diff
@@ -0,0 +1,19 @@
+package com.google.univiz.api;
+
+import java.util.ArrayList;
+import java.util.List;
+
+class SearchResourceImpl implements SearchResource {
+
+  @Override
+  public List<SearchData> getSearchSuggestions(String partialCollegeName) {
```
ihsan314, line 9: Make this static, since it doesn't seem instance-dependent.

mmurray22:
@ccondit no I don't believe it is ready for review right now because of the lack of testing. Working on that now!

capstone/src/main/java/com/google/univiz/api/SearchResource.java
```diff
@@ -18,5 +18,5 @@
    *     possible matches. The list is also not an exhaustive set of matches. It only includes few
    *     top results.
    */
-  List<SearchData> getSearchSuggestions(String collegeNames);
+  List<SearchData> getSearchSuggestions(String partialCollegeName);
```
ihsan314, line 21: Since you made it static in the implementation you should make it static here too.

capstone/src/main/java/com/google/univiz/api/SearchData.java
kaveh096: The CollegeData AutoValue has a field for id which is of type CollegeId. In your SearchResourceImpl class, you would need to copy the id from CollegeData into your SearchData representation. you can of course get the integer inside it or you can copy it wholesome into SearchData. Both works.

capstone/src/main/java/com/google/univiz/api/SearchResourceImpl.java
ihsan314: I guess it's now instance dependent, so you can leave it non-static.

capstone/src/main/java/com/google/univiz/api/SearchData.java
ccondit: My 2 cents: using `CollegeId` protects us from future change (ie we need a string instead of an int).

capstone/src/main/java/com/google/univiz/api/SearchResource.java
```diff
@@ -1,7 +1,22 @@
 package com.google.univiz.api;
 
+import java.util.List;
+
 /**
  * The SearchResoure interface is responsible for all data management/processing specific for the
  * search feature of the UniViz webapp.
  */
-public interface SearchResource {}
+public interface SearchResource {
```
ccondit, line 9: Wasn't this just reviews in #57? Mind updating the base of this PR to reflect that?

capstone/src/main/java/com/google/univiz/api/SuggestionData.java
```diff
@@ -0,0 +1,20 @@
+package com.google.univiz.api;
+
+/** SuggestionData represents the college data specific to the Search feature */
+public class SuggestionData {
```
ccondit, line 4: Why not use an @AutoValue?

capstone/src/main/java/com/google/univiz/api/SuggestionDataApi.java
```diff
@@ -11,5 +11,5 @@
    * Takes a fragment of a college name or a complete college name and returns a list of
    * SuggestionData objects
    */
-  List<SuggestionData> getCollegeSuggestions(String collegeName);
+  List<SuggestionData> getSuggestions(String collegeName);
```
ccondit, line 14: Why is this changing in this PR?

capstone/src/main/java/com/google/univiz/api/SearchResourceImpl.java
```diff
@@ -0,0 +1,24 @@
+package com.google.univiz.api;
+
+// TODO: imports
+//
```
ccondit, line 4: nit - this TODO format isn't correct, and doesn't seem to be needed...

capstone/src/main/java/com/google/univiz/api/SearchResourceImpl.java
```diff
@@ -0,0 +1,24 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+import javax.inject.Inject;
+
+class SearchResourceImpl implements SearchResource {
```
ccondit, line 9: nit - +final

capstone/src/test/java/com/google/univiz/api/SearchResourceImplTest.java
```diff
@@ -0,0 +1,36 @@
...
+import org.junit.runners.JUnit4;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnit;
+import org.mockito.junit.MockitoRule;
+
+@RunWith(JUnit4.class)
+public final class SearchResourceImplTest {
+  @Mock private SuggestionDataApi mockSuggestionApi;
+
+  @Rule public MockitoRule mockitoRule = MockitoJUnit.rule();
```
ccondit, line 20: nit - +final

capstone/src/test/java/com/google/univiz/api/SearchResourceImplTest.java
```diff
@@ -0,0 +1,36 @@
...
+
+  @Test
+  public void testGetSuggestions() throws Exception {
+    String partialCollegeName = "Sta";
+    SuggestionData stanford = new SuggestionData("Stanford University", 0);
+    List<SuggestionData> cannedResponse = new ArrayList<>();
+    cannedResponse.add(stanford);
```
ccondit, line 27: Leverage [Guava utilities](https://guava.dev/releases/23.0/api/docs/com/google/common/collect/Lists.html#newArrayList-E...-) for brevity.

capstone/src/test/java/com/google/univiz/api/SearchResourceImplTest.java
```diff
@@ -0,0 +1,36 @@
...
+    when(mockSuggestionApi.getSuggestions(partialCollegeName)).thenReturn(cannedResponse);
+
+    SearchResourceImpl search = new SearchResourceImpl(mockSuggestionApi);
+    List<SearchData> ret = search.getSearchSuggestions(partialCollegeName);
+    SearchData expected = SearchData.create("Stanford University", 0);
+    assertThat(expected.getCollegeName()).isEqualTo(ret.get(0).getCollegeName());
+    assertThat(expected.getCollegeId()).isEqualTo(ret.get(0).getCollegeId());
```
ccondit, line 34: This might read better as:
`assertThat(ret).containsExactly(expected);`

capstone/src/test/java/com/google/univiz/api/SearchResourceImplTest.java
```diff
@@ -0,0 +1,36 @@
...
+import org.mockito.junit.MockitoRule;
+
+@RunWith(JUnit4.class)
+public final class SearchResourceImplTest {
+  @Mock private SuggestionDataApi mockSuggestionApi;
+
+  @Rule public MockitoRule mockitoRule = MockitoJUnit.rule();
+
+  @Test
+  public void testGetSuggestions() throws Exception {
```
ccondit, line 23: Also - how about tests for corner cases:
* Empty lists?
* An exception being thrown?

capstone/src/main/java/com/google/univiz/api/SearchResourceImpl.java
```diff
@@ -0,0 +1,24 @@
...
+  SearchResourceImpl(SuggestionDataApi suggestionApi) {
+    this.suggestionApi = suggestionApi;
+  }
+
+  @Override
+  public List<SearchData> getSearchSuggestions(String partialCollegeName) {
+    List<SuggestionData> collegeSuggestions = suggestionApi.getSuggestions(partialCollegeName);
+    return collegeSuggestions.stream()
+        .map(college -> SearchData.create(college.collegeName(), college.collegeId()))
+        .collect(Collectors.toList());
```
kaveh096, line 22: too fancy

capstone/src/main/java/com/google/univiz/api/SearchResourceImpl.java
```diff
@@ -0,0 +1,24 @@
...
+// TODO: imports
+//
+import java.util.List;
+import java.util.stream.Collectors;
+import javax.inject.Inject;
+
+class SearchResourceImpl implements SearchResource {
+  private final SuggestionDataApi suggestionApi;
+
+  @Inject
```
kaveh096, line 12: guice is in the house

capstone/src/main/java/com/google/univiz/api/SuggestionData.java
kaveh096: +1

capstone/src/test/java/com/google/univiz/api/SearchResourceImplTest.java
```diff
@@ -0,0 +1,36 @@
...
+    String partialCollegeName = "Sta";
+    SuggestionData stanford = new SuggestionData("Stanford University", 0);
+    List<SuggestionData> cannedResponse = new ArrayList<>();
+    cannedResponse.add(stanford);
+    when(mockSuggestionApi.getSuggestions(partialCollegeName)).thenReturn(cannedResponse);
+
+    SearchResourceImpl search = new SearchResourceImpl(mockSuggestionApi);
```
kaveh096, line 30: I suspect you need to create this object in every test. So move this to be a field. Initialize it in the `@Before` method.

capstone/src/main/java/com/google/univiz/api/SuggestionData.java
```diff
@@ -0,0 +1,20 @@
...
+  private String name;
+  private int id;
```
kaveh096, line 6: Same argument about using `CollegeId`.

capstone/src/main/java/com/google/univiz/api/SearchResourceImpl.java
mmurray22: should I change the return haha

capstone/src/main/java/com/google/univiz/api/SearchResourceImpl.java
mmurray22: yay!

capstone/src/main/java/com/google/univiz/api/SearchResourceImpl.java
mmurray22: yes indeed! removed!

capstone/src/test/java/com/google/univiz/api/SearchResourceImplTest.java
mmurray22: added it to the SuggestionDataApi

capstone/src/main/java/com/google/univiz/api/SearchResourceImpl.java
mmurray22: addressed!

capstone/src/test/java/com/google/univiz/api/SearchResourceImplTest.java
mmurray22: hmm tried adding it and it broke the test...

ccondit: This comment is about line 20:
```suggestion
  @Rule public final MockitoRule mockitoRule = MockitoJUnit.rule();
```

mmurray22: ah that makes a lot more sense-thanks!

capstone/src/main/java/com/google/univiz/api/SearchResourceImpl.java
kaveh096: no

capstone/src/test/java/com/google/univiz/api/SearchResourceImplTest.java
mmurray22: I was going to make a follow-up PR, but I'll add those directly in on this PR :)

capstone/src/main/java/com/google/univiz/api/SuggestionDataApi.java
mmurray22: I will revert this change! I will be making another PR related to the SuggestionDataApi interface so I will include any other changes I want in that PR

capstone/src/main/java/com/google/univiz/api/SuggestionData.java
mmurray22: I decided not to do that in this PR since I just needed a simple prototype. However, I will be adding all the SuggestionData changes I made here: https://github.com/googleinterns/univiz/pull/84 to this PR

capstone/src/main/java/com/google/univiz/api/SearchResourceImpl.java
```diff
@@ -0,0 +1,31 @@
...
+  public SearchResourceImpl(SuggestionDataApi suggestionApi) {
+    this.suggestionApi = suggestionApi;
+  }
+
+  @Override
+  public List<SearchData> getSearchSuggestions(String partialCollegeName) {
+    if (partialCollegeName == null || !partialCollegeName.matches("[a-zA-Z]+")) {
```
ccondit, line 22: What if the client is trying to complete "Boston University" and types "Boston Uni"? Won't this regex preclude that?

capstone/src/main/java/com/google/univiz/api/SearchResourceImpl.java
```diff
@@ -0,0 +1,31 @@
...
+      return Lists.newArrayList(SearchData.create("", CollegeId.create(0)));
```
ccondit, line 23: How about returning an empty list instead of a list with a single dummy element in it. That's a clear signal to the client that there weren't any results, instead of requiring the client to inspect the list for a dummy result.

capstone/src/main/java/com/google/univiz/api/SearchResourceImpl.java
```diff
@@ -0,0 +1,31 @@
...
+    }
+    System.out.println(partialCollegeName);
```
ccondit, line 25: Let's not leave commit sysouts to the repo. If you want to log something, please use https://github.com/google/flogger

capstone/src/test/java/com/google/univiz/api/SearchResourceImplTest.java
```diff
@@ -0,0 +1,76 @@
...
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnit;
+import org.mockito.junit.MockitoRule;
+
+@RunWith(JUnit4.class)
+public final class SearchResourceImplTest {
+  @Mock private SuggestionDataApi mockSuggestionApi;
+  @Mock private SuggestionResponse cannedResponse;
```
ccondit, line 24: nit - since it's pretty lightweight to construct these objects, these tests would be more readable if you used some constant data instead of mocks. you can set them up once as constants in this class and that way every test method doesn't need to configure these mocks.

capstone/src/test/java/com/google/univiz/api/SearchResourceImplTest.java
```diff
@@ -0,0 +1,76 @@
...
+
+  @Before
+  public void setup() {
+    search = new SearchResourceImpl(mockSuggestionApi);
+  }
+
+  @Test
+  public void testSuggestionsForCollegeNameNoSpace() throws Exception {
+    String collegeName = "Stanford";
+    SearchResourceImpl search = new SearchResourceImpl(mockSuggestionApi);
```
ccondit, line 37: Isn't `search` already initialized in your `setup` method?

capstone/src/main/java/com/google/univiz/api/SearchResourceImpl.java
ccondit: Also, OOC - are we sure this is the right pattern to use? I can think of some counter examples:
* Texas A&M University
* University of California, Berkeley
* CO State-Pueblo
* University of Hawaiʻi at Mānoa 

... and so forth.

capstone/src/main/java/com/google/univiz/api/SearchResourceImpl.java
mmurray22: removed!

capstone/src/test/java/com/google/univiz/api/SearchResourceImplTest.java
mmurray22: yes indeed-removed!

capstone/src/main/java/com/google/univiz/api/SearchResourceImpl.java
mmurray22: is it better to remove the match statement altogether? seems like there are enough cases violating any reasonable regex that it doesn't make it work it

capstone/src/main/java/com/google/univiz/api/SearchResourceImpl.java
mmurray22: fixed!

capstone/src/test/java/com/google/univiz/api/SearchResourceImplTest.java
```diff
@@ -0,0 +1,75 @@
...
+    String nullCollegeName = null;
+    List<SearchData> ret = search.getSearchSuggestions(nullCollegeName);
+    CollegeId collegeId = CollegeId.create(0);
+    SearchData expected = SearchData.create("", collegeId);
+    assertThat(ret).containsExactly(expected);
+  }
+
+  @Test
+  public void testSuggestionsForInvalidSuggestion() throws Exception {
+    String invalidCollegeName = "%$#@!1235566";
```
kaveh096, line 69: I don't think you need this test case now that you have removed the regex check.

Issue #38, opened 2020-07-31
Suggested Timeline for College Applications


ihsan314:
Unfortunately, the US government CollegeScoreCard API does not have information on application deadlines for the various colleges. But one of UniViz's goals is to inform students of application deadlines in a visual format. Until we find a high quality source on application deadlines, UniViz will present suggested deadlines to the user, based on how many colleges the user has chosen. This will be in the form of a frequency map that looks something like the following:

![Frequency Map](https://user-images.githubusercontent.com/43079005/89064698-531bd000-d31f-11ea-9af3-6f7dbe73f359.png)

Additionally we can have something like the following:

![Timeline](https://user-images.githubusercontent.com/43079005/89065392-69765b80-d320-11ea-8561-0b5a81151d33.png)

This, instead of providing a frequency map, would actually provide the user of a suggested timeline to complete applications with minimal stress.

Both styles will use the free Google Charts API. Data wise, a Java servlet will have a hardcoded JSON object consisting of suggested timelines. A JavaScript file will fetch this JSON object and convert it to appropriate [DataTable](https://developers-dot-devsite-v2-prod.appspot.com/chart/interactive/docs/reference#DataTable) objects for each type of chart.

PR #61, opened 2020-08-06
Set up visualization interface and datatypes

ihsan314:
Iterate on all of #23, #26, and #38 by preparing the visualization datatypes and the format of the `VizResource` interface.

I should probably make the classes `AutoValues` since I'm likely to serialize these into JSON objects in my servlet. Additionally I should probably deal with Guice Integration. Before I do that, I want to catch other potential issues with my layout first.


capstone/src/main/java/com/google/univiz/api/VisResource.java
```diff
@@ -1,7 +1,19 @@
...
  * feature of the UniViz webapp.
  */
-public interface VisResource {}
+public interface VisResource {
+
+  List<TuitionStats> getTuitionStats(List<String> colleges);
+
+  List<Timeline> getRecommendedTimeline(List<String> colleges);
+
+  List<Deadline> getDeadlines(List<String> colleges);
```
biancamacias, line 18: wouldnt these all be taking in a ```List<CollegeData>```? i think it would, because that is the type that is going to have the information you want in the type that you want it.

ihsan314: Ah ok I will change the methods to take in a CollegeData. Though I thought there was going to be something more [lightweight](https://github.com/googleinterns/univiz/pull/52#discussion_r465224050) seeing Michaela's suggestion and the [endorsements](https://github.com/googleinterns/univiz/pull/52#discussion_r465400360) of Chris and Kaveh. I had the impression that these methods would take in the list of colleges and then use CollegeData to fetch the JSON and then populate the appropriate data structures. But I guess it works if I directly take in a CollegeData object too.

ihsan314:
> just a note -- i think you're right about `AutoValue` for all of these classes, `Deadline`, `Timeline`, and `TuitionStats`, but i don't think you'll be doing any serializing -- that will be done already for you as #52 is taking care of the serializing and parsing the JSON into `CollegeData` types.
> 
> take a look at the `ScorecardData.java` class in #37 for `AutoValue` reference.

Ah ok thanks for confirming my suspicions about `AutoValue`, I'll review #37.

I thought serializing would come into play when considering that as the output of the servlet, I'll need a JSON, and especially in the case of `Deadline` and `Timeline`, they will likely not correspond to any fields in `CollegeData`. Additionally for `TuitionStats` I may consider including results computed based on fields in `CollegeData`, so there may be additional disparity there. We can discuss this more as this and #52 progress.

capstone/src/main/java/com/google/univiz/api/VisResource.java
```diff
@@ -1,7 +1,20 @@
 package com.google.univiz.api;
```
mmurray22, line 1: Should I cease work on the VisResource Interface?

capstone/src/main/java/com/google/univiz/api/dataviz/Deadline.java
```diff
@@ -0,0 +1,3 @@
+package com.google.univiz.api.dataviz;
+
```
mmurray22, line 2: add JavaDoc event to event empty objects

ihsan314: Ah good catch, I overlooked that. Will do.

biancamacias:
> > just a note -- i think you're right about `AutoValue` for all of these classes, `Deadline`, `Timeline`, and `TuitionStats`, but i don't think you'll be doing any serializing -- that will be done already for you as #52 is taking care of the serializing and parsing the JSON into `CollegeData` types.
> > take a look at the `ScorecardData.java` class in #37 for `AutoValue` reference.
> 
> Ah ok thanks for confirming my suspicions about `AutoValue`, I'll review #37.
> 
> I thought serializing would come into play when considering that as the output of the servlet, I'll need a JSON, and especially in the case of `Deadline` and `Timeline`, they will likely not correspond to any fields in `CollegeData`. Additionally for `TuitionStats` I may consider including results computed based on fields in `CollegeData`, so there may be additional disparity there. We can discuss this more as this and #52 progress.

ohhhh i see. yes, youll have to convert the object to JSON when you send it to the front end, but you wont have to do from JSON to java object is what i meant because that is what #52 will do, and we can just use the ```CollegeData``` class to get what we need.

capstone/src/main/java/com/google/univiz/api/VisResource.java
ihsan314: I believe you no longer need to work on `VisResource`, so yes.

capstone/src/main/java/com/google/univiz/api/VisResource.java
biancamacias: `CollegeData` isn't going to be involved in any parsing, i dont believe. but we can talk about this offline.

capstone/src/main/java/com/google/univiz/api/VisResource.java
```diff
@@ -1,7 +1,20 @@
...
+import java.util.List;
+
 /**
  * The VisResource interface is responsible for all data management/processing specific for the vis
  * feature of the UniViz webapp.
  */
-public interface VisResource {}
+public interface VisResource {
+
+  List<TuitionStats> getTuitionStats(List<CollegeData> colleges);
```
mmurray22, line 15: Each fxn in the interface should have it's own javadoc

ihsan314: Ah good catch, I forget to deal with JavaDoc.

capstone/src/main/java/com/google/univiz/api/VisResource.java
```diff
@@ -1,7 +1,40 @@
...
+
+  /**
+   * For every college the user selected, obtain the recommended set of actions for the user to take
+   * and the recommended time range for each action.
+   *
+   * @param colleges the raw data for each of the user's selected colleges
+   * @return the actions/time range pairs for each college.
+   */
+  List<Timeline> getRecommendedTimeline(List<CollegeData> colleges);
```
ccondit, line 31: What's the difference between deadline and timeline? They seem very similar. Is timeline a collection of deadlines?

capstone/src/main/java/com/google/univiz/api/VisResource.java
```diff
@@ -1,7 +1,40 @@
...
+
+  /**
+   * For every college the user selected, obtain the application deadlines.
+   *
+   * @param colleges the raw data for each of the user's selected colleges
+   * @return the application deadlines (application release/due date) per college.
+   */
+  List<Deadline> getDeadlines(List<CollegeData> colleges);
```
ccondit, line 39: OOC where is this timeline data going to come from?

capstone/src/main/java/com/google/univiz/api/VisResource.java
```diff
@@ -1,7 +1,40 @@
...
+public interface VisResource {
+
+  /**
+   * For every college the user selected, obtain the raw statistics and compute summary statistics
+   * on tuition-related data.
+   *
+   * @param colleges the raw data for each of the user's selected colleges
+   * @return the list of statistics per college.
+   */
+  List<TuitionStats> getTuitionStats(List<CollegeData> colleges);
```
ccondit, line 22: Is the client going to have to supply all of the college data back to the methods in this resource? Or can the client simply supply the list of CollegeIds? 

One corner case I can think of is the first load, when the client doesn't have any access to collegedata...

ihsan314: I think the methods can still behave as intended even if `CollegeIds` are passed. I will change the signature for these methods.

capstone/src/main/java/com/google/univiz/api/VisResource.java
ihsan314: A timeline ideally would have a range of deadlines associated with the various steps of completing an application. For instance an example timeline could have all of the following included:
- contacting potential application references for the first three weeks of the application period.
- writing the essay for the next three weeks.
- researching in depth into the university for the next two weeks.
whereas a deadline type would consist of application start/end dates only.

capstone/src/main/java/com/google/univiz/api/VisResource.java
ihsan314: The deadline and timeline data are going to be hardcoded for now. The deadline data ideally wouldn't be hardcoded but there is currently no suitable API for it. The timeline data will consist of predefined generic actions that apply to most colleges, along with a particular number of weeks per action. Ideally if there was a suitable API for application deadlines, then the timeline data would contain dates relevant to the specific colleges. But for now the deadlines dates will be hardcoded.

capstone/src/main/java/com/google/univiz/api/VisResource.java
kaveh096: +1 to passing in CollegeIds only.

capstone/src/main/java/com/google/univiz/api/VisResource.java
```diff
@@ -1,7 +1,39 @@
...
  * feature of the UniViz webapp.
  */
-public interface VisResource {}
+public interface VisResource {
+
+  /**
+   * For every college the user selected, obtain the raw statistics and compute summary statistics
+   * on tuition-related data.
+   *
+   * @param colleges the raw data for each of the user's selected colleges
```
kaveh096, line 18:
```suggestion
   * @param colleges the IDs for each of the user's selected colleges
```

capstone/src/main/java/com/google/univiz/api/VisResource.java
```diff
@@ -1,7 +1,39 @@
...
+   * @return the list of statistics per college.
+   */
+  List<TuitionStats> getTuitionStats(List<CollegeId> colleges);
+
+  /**
+   * For every college the user selected, obtain the recommended set of actions for the user to take
+   * and the recommended time range for each action.
+   *
+   * @param colleges the raw data for each of the user's selected colleges
```
kaveh096, line 27:
```suggestion
   * @param colleges the IDs for each of the user's selected colleges
```

capstone/src/main/java/com/google/univiz/api/VisResource.java
```diff
@@ -1,7 +1,39 @@
...
+   * @return the actions/time range pairs for each college.
+   */
+  List<Timeline> getRecommendedTimeline(List<CollegeId> colleges);
+
+  /**
+   * For every college the user selected, obtain the application deadlines.
+   *
+   * @param colleges the raw data for each of the user's selected colleges
```
kaveh096, line 35:
```suggestion
   * @param colleges the IDs for each of the user's selected colleges
```

PR #65, opened 2020-08-07
Add fixing/replacing option to linters and formatters

ihsan314:
I've had enough of manually running the formatters before committing, so I've added the `--fix` option to the JS/CSS/Markdown linters and added `--replace` to the Java formatter. Not only that the linters will commit the progress they made in reformatting the files, so for most things it should hopefully be very helpful.

I haven't really tested the modifications to the four linters much, aside from committing to this branch and making sure they behave properly when there are no diffs.

### DIFF
--- a/.github/workflows/csslint.yaml
+++ b/.github/workflows/csslint.yaml
@@ -17,4 +17,4 @@ jobs:
           git config user.name "CSS Presubmit"
           git add .
           git commit -m "Format CSS files"
-          git push
+          git push origin `git rev-parse --abbrev-ref HEAD`
--- a/.github/workflows/csslint.yaml
+++ b/.github/workflows/csslint.yaml
@@ -17,5 +17,5 @@ jobs:
           git config user.name "CSS Presubmit"
           git add .
           if [ git commit -m "Format CSS files" ]; then
-            git push origin `git rev-parse --abbrev-ref HEAD`
+            git push origin `git branch --show-current`
           fi
--- a/.github/workflows/csslint.yaml
+++ b/.github/workflows/csslint.yaml
@@ -12,7 +12,7 @@ jobs:
         run: npm install
       - name: Lint
         # Run StyleLint on diffed CSS files. Do not run if there are no diffs.
-        run:
+        run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/javalint.yaml
```diff
@@ -19,10 +19,12 @@ jobs:
         run: |
           for file in $(git diff --diff-filter=ACM --name-only origin/master HEAD -- . ':!node_modules' | grep -E "(.*)\.(java)$"); do
             echo -e "Running google-java-format on:\n $file\n"
-            if ! java -jar ./google-java-format-1.8-all-deps.jar $file | diff $file -; then
-              EXIT=1
-            fi
+            java -jar ./google-java-format-1.8-all-deps.jar $file > $file
```
ccondit, line 22:
```suggestion
            java -jar ./google-java-format-1.8-all-deps.jar --replace $file
```

.github/workflows/csslint.yaml
```diff
@@ -12,4 +12,10 @@ jobs:
         run: npm install
       - name: Lint
         # Run StyleLint on diffed CSS files. Do not run if there are no diffs.
-        run: git diff --diff-filter=ACM --name-only origin/master HEAD -- . ':!node_modules' | grep -E "(.*)\.(css|htm|html)$" | xargs --no-run-if-empty npx stylelint
+        run: |
+          git diff --diff-filter=ACM --name-only origin/master HEAD -- . ':!node_modules' | grep -E "(.*)\.(css|htm|html)$" | xargs --no-run-if-empty npx stylelint --fix
+          git config user.name "CSS Presubmit"
+          git add .
+          if [ git commit -m "Format CSS files" ]; then
+            git push origin `git rev-parse --abbrev-ref HEAD`
```
ccondit, line 20: Will this work? I think it's a bit clearer. If so, can you change the other files, as well?

```suggestion
            git push origin `git branch --show-current`
```

ihsan314: It is clearer, but only works in version 2 of git, which might not be installed. At least it's not installed locally.

ihsan314: I'll try it and if it fails to work I'll revert it to the less clear version.

PR #64, opened 2020-08-07
Converter than converts ScorecardData.java class into CollegeData.java class

biancamacias:
This PR is about the [Converter](https://github.com/google/guava/blob/master/guava/src/com/google/common/base/Converter.java) that will make the `ScorecardData` object into a `CollegeData` object. It also includes adding the `CarnegieSizeDegree.java` class, modifying the tests for all classes, and removing the `urbanizationDegree` from all classes due to its continuous `null` values from the API. This PR also adds a builder to the `ScorecardData.java` class.

PR #66, opened 2020-08-07
CollegeData.java class 

biancamacias:
This PR is specifically for the class `CollegeData` which will be used in PR #52 interface. The `CollegeData` class currently does not have a getter method, but one will be added once #64 is merged into this PR. `CollegeId` is a dummy class, the implementation of this class might change later.

The test `CarnegieSizeDegreeTest.java` tests to make sure that indexing in the `CarnegieSizeDegree` `getDegree` method is not randomized.


biancamacias:
> Was this reviewed in #64?

yes. michaela thought it would be better if i moved this class to another PR, and im addressing the changes you requested from #64 here. ill re-request your review when ive made those changes, which includes writing some tests. :)

capstone/src/main/java/com/google/univiz/CarnegieSizeDegree.java
```diff
@@ -0,0 +1,33 @@
+package com.google.univiz;
+
+/**
+ * CarnegieSizeDegree is a class that holds the enum for the range of values for a school's carnegie
+ * size degree.
+ */
+public enum CarnegieSizeDegree {
+  NOT_AVAILABLE("Degree unavailable"),
+  SIZE1("Two-year college, very small"),
```
kaveh096, line 9: Any reason not to use more descriptive names for these enum constants? For example `SIZE1` could be renamed to `TWO_YEAR_COLLEGE`.

kaveh096, line 9: I remember you had another problem with trying to convert integer numbers returned from scorecard REST api into these enums. Since you're now using constrctor arguments to capture the humand-readable text for these enums, you can also add another argument and capture the numeric representation of these enums:

```java
public enum CarnegieSizeDegree {
  TWO_YEAR_COLLEGE("Two-year college, very small", 1),
   //...
  private final int scorecardNumericValue;
  private final String degree;

  CarnegieSizeDegree(String degree, int scorecardNumericValue) {
    this.degree = degree;
    this.scorecardNumericValue = scorecardNumericValue;
  }

  static CarnegieSizeDegree fromNumericValue(int scorecardNumericValue) {
    for (CarnegieSizeDegree current : values()) {
      if (current.scorecardNumericValue == scorecardNumericValue) { return current; }
    }
    throw new OutOfRangeException(
        String.format("Enum numeric value not found: %d", scorecardNumericValue));
  }
```

capstone/src/main/java/com/google/univiz/CollegeData.java
```diff
@@ -0,0 +1,85 @@
...
+
+  /**
+   * The CollegeData builder has some set default values. The avgSat and admissionRate values are
+   * estimated averages as of 2020.
+   */
+  static Builder builder() {
+    return new AutoValue_CollegeData.Builder()
+        .setCarnegieSizeDegree(CarnegieSizeDegree.NOT_AVAILABLE)
+        .setAvgSat(1059.0)
+        .setAdmissionRate(0.65);
```
ccondit, line 51: What are the meaning of these defaults? Can't we just omit them?

capstone/src/main/java/com/google/univiz/CarnegieSizeDegree.java
biancamacias: should i do this as opposed to what i do in [this file](capstone/src/main/java/com/google/univiz/CarnegieSizeDegree.java) from #64?

capstone/src/main/java/com/google/univiz/CollegeData.java
biancamacias: these are averages i found. i used these because i didnt want it to say that the admission rate would be 0 or average sat would be 0 for a particular college.

capstone/src/main/java/com/google/univiz/CollegeId.java
```diff
@@ -0,0 +1,10 @@
+package com.google.univiz;
+
+import com.google.auto.value.AutoValue;
+
+/** CollegeId is a dummy class just to override compilation errors. */
+@AutoValue
+abstract class CollegeId {
+
+  abstract int id();
```
mmurray22, line 9: will collegeId only have ids or will there also be a string attached?

biancamacias: not sure. thats up to whoever is using the class, i wasnt sure what it needed.

capstone/src/main/java/com/google/univiz/CarnegieSizeDegree.java
kaveh096: Either way is fine. As long as you add tests to make sure it works.

capstone/src/test/java/com/google/univiz/CarnegieSizeDegreeTest.java
```diff
@@ -0,0 +1,20 @@
...
+@RunWith(JUnit4.class)
+public final class CarnegieSizeDegreeTest {
+
+  @Test
+  public void testIndexing() {
+    assertThat(CarnegieSizeDegree.NOT_AVAILABLE).isEqualTo(CarnegieSizeDegree.getDegree(0));
+    assertThat(CarnegieSizeDegree.VSMALL_NONRESI_FOUR_YEAR_COLLEGE)
+        .isEqualTo(CarnegieSizeDegree.getDegree(6));
+    assertThat(CarnegieSizeDegree.SMALL_TWO_YEAR_COLLEGE)
+        .isEqualTo(CarnegieSizeDegree.getDegree(2));
```
ccondit, line 18: Just wanted to make a note that typically asserts are written like:
`assertThat(actual).isEqualtTo(expected)`.

You'll get a lot more support from the fluent assertions like this, especially when you get into lists:
`assertThat(actual).contains("foo")`

biancamacias:
> Weren't these just reviewed in #64?

yes, the base branch was this branch, and so it merged with this one ;(

it was my fault, i should've checked before merging. but its all the same code, it just overrode what was in here.

PR #67, opened 2020-08-07
Adds Suggestion API functions

mmurray22:
Interface for the Suggestion API Interface which will get the suggested college data for the search feature by communicating with the REST API.


capstone/src/main/java/com/google/univiz/SuggestionAPIData.java
```diff
@@ -0,0 +1,15 @@
...
+/**
+ * Communicates directly with the REST API to get a list of college suggestions related to a client
+ * query
+ */
+public interface SuggestionDataAPI {
+  /**
+   * Takes a list of college names and returns a list of SuggestionData objects, where each object
+   * is unique to each college name providede in the first list
+   */
+  List<SuggestionData> getCollegeSuggestions(List<String> collegeNames);
```
ihsan314, line 14: Chris pointed this out earlier, but are we expecting a list of college names from the client? Does it make sense to expect this?

capstone/src/main/java/com/google/univiz/SuggestionAPIData.java
```diff
@@ -0,0 +1,15 @@
+package com.google.univiz;
+
+import java.util.List;
+
+/**
+ * Communicates directly with the REST API to get a list of college suggestions related to a client
+ * query
+ */
+public interface SuggestionDataAPI {
```
biancamacias, line 9:
```suggestion
public interface SuggestionDataApi {
```

biancamacias:
i have a feeling youre getting build errors because of the class `SuggestionData` not existing? but not im not sure about the `UnivizConfig` file not working with `AutoValue`? that is really strange.

capstone/src/main/java/com/google/univiz/SuggestionAPIData.java
mmurray22: fixed this in the other PR, will update it here :)

capstone/src/main/java/com/google/univiz/SuggestionData.java
```diff
@@ -0,0 +1,4 @@
+package com.google.univiz;
+
+/** SuggestionData represents the college data specific to the Search feature */
+public class SuggestionData {}
```
ihsan314, line 4: Would we benefit from using `AutoValue` for this?

mmurray22: yes :) However since it is just a placeholder here, I did not incorporate AutoValue in this PR

capstone/src/main/java/com/google/univiz/SuggestionDataApi.java
```diff
@@ -0,0 +1,15 @@
...
+
+import java.util.List;
+
+/**
+ * Communicates directly with the REST API to get a list of college suggestions related to a client
+ * query
+ */
+public interface SuggestionDataApi {
+  /**
+   * Takes a list of college names and returns a list of SuggestionData objects, where each object
```
ccondit, line 11: Can you true up this comment to match the method?
* No longer a list argument.
* Also s/providede/provided/.

capstone/src/main/java/com/google/univiz/SuggestionDataApi.java
```diff
@@ -0,0 +1,15 @@
+package com.google.univiz;
+
+import java.util.List;
+
+/**
+ * Communicates directly with the REST API to get a list of college suggestions related to a client
```
ccondit, line 6: Avoid implementation details in the interface documentation. Today we'll use the REST API but tomorrow it could be a database or something else entirely. If this class stays agnostic to the implementation then the documentation won't need updating for situations like that...

capstone/src/main/java/com/google/univiz/SuggestionDataApi.java
mmurray22: updated!

capstone/src/main/java/com/google/univiz/SuggestionDataApi.java
mmurray22: I updated the documentation. However, on a related note-should I still be communicating directly with the REST API in the implementation of this interface  (in private functions)?

capstone/src/main/java/com/google/univiz/SuggestionDataApi.java
```diff
@@ -3,13 +3,13 @@
...
- * query
+ * Communicates directly with the data source to get a list of college suggestions related to a
+ * client query
  */
 public interface SuggestionDataApi {
   /**
-   * Takes a list of college names and returns a list of SuggestionData objects, where each object
-   * is unique to each college name providede in the first list
+   * Takes a college name and returns a list of SuggestionData objects, where each object is unique
+   * to each college name provided in the first list
```
ccondit, line 12: what does "provided in the first list" mean, in this case?

capstone/src/main/java/com/google/univiz/SuggestionDataApi.java
ccondit: > should I still be communicating directly with the REST API in the implementation of this interface (in private functions)?

Oh yes! And that implementation will be in the `scorecard` package where you can feel free to discuss the specifics of the implementation in JavaDoc :)

capstone/src/main/java/com/google/univiz/SuggestionDataApi.java
mmurray22: updated!

capstone/src/main/java/com/google/univiz/SuggestionDataApi.java
```diff
@@ -7,9 +7,6 @@
  * client query
  */
 public interface SuggestionDataApi {
-  /**
-   * Takes a college name and returns a list of SuggestionData objects, where each object is unique
-   * to each college name provided in the first list
-   */
+  /** Takes a college name and returns a list of SuggestionData objects */
```
ccondit, line 10: Might be good to be clear on what `collegeName` can be:

```suggestion
  /** Takes a fragment of a college name or a complete college name and returns a list of SuggestionData objects */
```

capstone/src/main/java/com/google/univiz/SuggestionDataApi.java
```diff
@@ -0,0 +1,15 @@
+package com.google.univiz;
+
+import java.util.List;
+
+/**
+ * Communicates directly with the data source to get a list of college suggestions related to a
+ * client query
+ */
+public interface SuggestionDataApi {
```
kaveh096, line 9: I'm confused, how is this interface different from [SearchResource](https://github.com/googleinterns/univiz/pull/57/files#diff-32b6eaa160ada5e3baf248f4e32d1a20R9) that you're working on in PR #57?

mmurray22: SearchResource is about translating the Generated Object data into a usable data form the server can use. SuggestionAPI talks directly to the REST API and gets the data from the source

kaveh096: In that case how is this interface different from [CollegeDataApi](https://github.com/googleinterns/univiz/blob/master/capstone/src/main/java/com/google/univiz/CollegeDataApi.java) class?

kaveh096: OK, after chatting with @mmurray22 offline, i now understand what is happening here. Looks like we have a parallel pipeline for auto complete feature. I'm not a big fan of this design but that is just my subjective opinion. Perhaps the implementation of these two interfaces would share some code.

Also both of these interfaces are right now in the root package while i thought they should be in the scorecard package. But again it's something we can come back and refactor later.

mmurray22: I can definitely put this in its own package. I was thinking once this pipeline is done, then we could think about making the two more cohesive. Let me know if you have any more questions!

PR #69, opened 2020-08-07
Test class with terrible formatting

ihsan314:
See if the Java formatting linter will auto commit a fix for the Dummy.java file.

The Dummy.java file should be changed to use two spaces instead of tabs, the imports should be alphabetized, the newline at the end of the file should be removed, and the long line should be broken up.


ihsan314:
Never mind there's still work to do, it's probably not committing to the right branch.

ihsan314:
What a lost cause.

PR #73, opened 2020-08-10
CollegeDataApi interface

biancamacias:
This is an interface for the `CollegeDataApiImpl.java` class, which is meant to be the only class that interacts with the CollegeScorecard API.

This interface has only one method, `getCollegesById`, which is meant to send the REST request into the backend and use the converter to convert from `ScorecardData` class to `CollegeData` class.

### DIFF
--- a/capstone/src/main/java/com/google/univiz/CollegeDataApi.java
+++ b/capstone/src/main/java/com/google/univiz/CollegeDataApi.java
@@ -1,13 +1,11 @@
 package com.google.univiz;
 
 import java.util.List;
-import java.util.Set;
 
 /**
- * CollegeDataApi is an interface for the CollegeDataApiImpl class that interacts with external APIs
- * to send REST requests.
+ * Provides access to college related data.
  */
 public interface CollegeDataApi {
 
-  List<CollegeData> getCollegesById(Set<CollegeId> ids);
+  List<CollegeData> getCollegesById(List<CollegeId> ids);
 }
--- a/capstone/src/main/java/com/google/univiz/CollegeDataApi.java
+++ b/capstone/src/main/java/com/google/univiz/CollegeDataApi.java
@@ -2,9 +2,7 @@ package com.google.univiz;
 
 import java.util.List;
 
-/**
- * Provides access to college related data.
- */
+/** Provides access to college related data. */
 public interface CollegeDataApi {
 
   List<CollegeData> getCollegesById(List<CollegeId> ids);
--- a/capstone/src/main/java/com/google/univiz/CollegeDataApi.java
+++ b/capstone/src/main/java/com/google/univiz/CollegeDataApi.java
@@ -1,0 +1,9 @@
+package com.google.univiz;
+
+import java.util.List;
+import java.util.Set;
+
+public interface CollegeDataApi {
+
+  List<CollegeData> getCollegesById(Set<CollegeId> ids);
+}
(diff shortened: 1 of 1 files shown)


capstone/src/main/java/com/google/univiz/CollegeDataApi.java
```diff
@@ -0,0 +1,13 @@
+package com.google.univiz;
+
+import java.util.List;
+import java.util.Set;
+
+/**
+ * CollegeDataApi is an interface for the CollegeDataApiImpl class that interacts with external APIs
+ * to send REST requests.
```
ccondit, line 8: I would encourage leaving out implementation details in interface javadoc. In particular, the name of the implementing class (which could change) and the means by which the implementation will fetch data (REST requests). Such details can quickly become stale.

This javadoc could be simply:

/** Provides access to college related data. */

capstone/src/main/java/com/google/univiz/CollegeDataApi.java
```diff
@@ -0,0 +1,13 @@
...
+ */
+public interface CollegeDataApi {
+
+  List<CollegeData> getCollegesById(Set<CollegeId> ids);
```
ccondit, line 12: I think there's an impedance mismatch here between the `Set` of ids and the returned `List`. As a reader of this signature, it's unclear how the results will be ordered. Do you want to change the returned `List` to be a `Set`, as well?

biancamacias: i dont think so. ill change it to a `List`

kaveh096: I actually liked the set input better because it clearly communicates that this API does not like to get duplicate IDs. But i also agree that a set input should be followed by a set output. But returning a list of better than set for output because we like to have stuff returned in some deterministic order. Otherwise college names could keep jumping around every time the user refreshes the page. No action item for now. Let's keep both as a list.

capstone/src/main/java/com/google/univiz/CollegeDataApi.java
```diff
@@ -0,0 +1,9 @@
+package com.google.univiz;
+
+import java.util.List;
+
+/** Provides access to college related data. */
+public interface CollegeDataApi {
```
kaveh096, line 6: It looks like other api interfaces are using the `{FeatureName}Resource` naming convention. Do we want to use the same convention for this interface too: `CollegeDataResource` or this is different because it's not tied to a particular feature?

capstone/src/main/java/com/google/univiz/CollegeDataApi.java
biancamacias: ok. we can talk about this at standup.

capstone/src/main/java/com/google/univiz/CollegeDataApi.java
biancamacias: not sure. it isn't tied to a particular feature, i think its just responsible for REST requests.

ccondit: IIUC the `**Api` classes are to be implemented inside the scorecard package. The `**Resource` classes will inject the `**Api` classes, and the servlet will inject the ``**Resource` classes. So I think this is ok to stay as `CollegeDataApi`. 

Does that match what everyone else is expecting?

biancamacias: yes i think so.

PR #76, opened 2020-08-12
CollegeScorecardAPI "metadata" and "results" parser

biancamacias:
This PR includes the `ScorecardResponse.java` class, the `ScorecardResponseTest.java` class which tests to make sure the JSON is parsed correctly, and the `scorecard_response.json` file which holds the raw JSON provided when calling the CollegeScorecardAPI. The `ScorecardResponse.java` class only extracts the `results` array, which is the only information needed for the UniViz app. 

When coding the `CollegeDataApiImpl.java` class, the `ScorecardReponse.java` class will be used to parse the REST JSON response easily. The method will loop through the `scorecardData()` and convert each `ScorecardData` object to a `CollegeData` object.

### DIFF
--- a/capstone/src/test/java/com/google/univiz/scorecard/ScorecardResponseTest.java
+++ b/capstone/src/test/java/com/google/univiz/scorecard/ScorecardResponseTest.java
@@ -1,0 +1,46 @@
+package com.google.univiz.scorecard;
+
+import static com.google.common.truth.Truth.assertThat;
+
+import com.google.common.io.Resources;
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+import com.ryanharter.auto.value.gson.GenerateTypeAdapter;
+import java.io.InputStreamReader;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public final class ScorecardResponseTest {
+
+  /**
+   * JSON from scorecard.json comes directly from the CollegeScorecard API.
+   *
+   * @see <a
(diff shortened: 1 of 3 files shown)


capstone/src/test/java/com/google/univiz/scorecard/ScorecardResponseTest.java
```diff
@@ -0,0 +1,46 @@
...
+   */
+  @Test
+  public void testJsonDeserializes() throws Exception {
+    Gson gson = new GsonBuilder().registerTypeAdapterFactory(GenerateTypeAdapter.FACTORY).create();
+    InputStreamReader scorecardReader =
+        new InputStreamReader(
+            Resources.getResource(ScorecardResponseTest.class, "scorecard_response.json")
+                .openStream());
+    ScorecardResponse response = gson.fromJson(scorecardReader, ScorecardResponse.class);
+    ScorecardData scorecardData = response.scorecardData().get(0);
```
ccondit, line 31: I think a useful assertion here would be:
```suggestion
    assertThat(response.scorecardData()).hasSize(1);
    ScorecardData scorecardData = response.scorecardData().get(0);
```

capstone/src/test/java/com/google/univiz/scorecard/ScorecardResponseTest.java
```diff
@@ -0,0 +1,46 @@
...
+    assertThat(scorecardData.flagMainCampus()).isEqualTo(1);
+    assertThat(scorecardData.latitude()).isEqualTo(40.729452);
+    assertThat(scorecardData.longitude()).isEqualTo(-73.997264);
+    assertThat(scorecardData.carnegieSizeDegree()).isEqualTo(17);
+    assertThat(scorecardData.admissionRate()).isEqualTo(0.1999);
+    assertThat(scorecardData.avgSat()).isEqualTo(1419.0);
+    assertThat(scorecardData.numOfUndergrads()).isEqualTo(26339);
+    assertThat(scorecardData.avgCost()).isEqualTo(69830);
+    assertThat(scorecardData.ratioOfMen()).isEqualTo(0.4253);
+    assertThat(scorecardData.ratioOfWomen()).isEqualTo(0.5747);
```
ccondit, line 44: Since this is already asserted on in the `ScorecardDataTest`, I think just asserting on one field would suffice...

capstone/src/test/java/com/google/univiz/scorecard/ScorecardResponseTest.java
biancamacias: that makes more sense.

PR #77, opened 2020-08-13
Organizing packages

biancamacias:
This PR is to address the file organization issues. Originally this PR was to adjust where `CollegeId` was going to be, but now I've moved lots of other classes into new, better organized packages. The structure will look like this:

- package com.google.univiz.api:
SuggestionDataApiImpl.java
SuggestionDataApi.java
CollegeDataApi.java
CollegeDataApiImpl.java

- package com.google.univiz.api.resource:
SearchResource.java
MapResource.java
VisResource.java
SearchResourceImpl.java
MapResourceImpl.java
VisResourceImpl.java

- package com.google.univiz.api.representation:
CollegeId.java
MapData.java
VisData.java
CollegeData.java
SuggestionData.java
CarnegieSizeDegree.java

- package com.google.univiz.scorecard:
ScorecardData.java
ScorecardConverter.java

PR #81, opened 2020-08-13
VisResource.java Class Set-Up

biancamacias:
This PR is to set up all the classes Ihsan will need to implement the `VisResource.java` interface. I made a converter to convert from `CollegeData` to `CollegeStats`. I also added the `CollegeDataDatavizConverterTest.java` class to make sure everything compiles and tests as I expect them to.

### DIFF
--- a/capstone/src/main/java/com/google/univiz/api/VisResource.java
+++ b/capstone/src/main/java/com/google/univiz/api/VisResource.java
@@ -1,8 +1,8 @@
 package com.google.univiz.api;
 
+import com.google.univiz.api.dataviz.CollegeStats;
 import com.google.univiz.api.dataviz.Deadline;
 import com.google.univiz.api.dataviz.Timeline;
-import com.google.univiz.api.dataviz.TuitionStats;
 import java.util.List;
 
 /**
@@ -13,12 +13,12 @@ public interface VisResource {
 
   /**
    * For every college the user selected, obtain the raw statistics and compute summary statistics
-   * on tuition-related data.
+   * on college-related data.
    *
    * @param colleges the IDs for each of the user's selected colleges
    * @return the list of statistics per college.
    */
-  List<TuitionStats> getTuitionStats(List<CollegeId> colleges);
+  List<CollegeStats> getCollegeStats(List<CollegeId> colleges);
 
   /**
    * For every college the user selected, obtain the recommended set of actions for the user to take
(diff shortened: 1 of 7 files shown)


capstone/src/main/java/com/google/univiz/api/dataviz/CollegeStats.java
```diff
@@ -0,0 +1,42 @@
+package com.google.univiz.api.dataviz;
+
+import com.google.auto.value.AutoValue;
+
+/** CollegeStats is a class that contains statistics related to tuition for a particular college. */
+@AutoValue
```
ccondit, line 6: nit - This AutoValue will be shipped to the client - so it will make sense to add a `@GenerateTypeAdapter`...

capstone/src/main/java/com/google/univiz/api/dataviz/CollegeDataDatavizConverter.java
```diff
@@ -0,0 +1,26 @@
+package com.google.univiz.api.dataviz;
+
+import com.google.common.base.Converter;
+import com.google.univiz.CollegeData;
+
+final class CollegeDataDatavizConverter extends Converter<CollegeData, CollegeStats> {
```
kaveh096, line 6: Here are some suggested names for this class:
CollegeDataToDataVizConverter
CollegeDataToCollegeStatsConverter
CollegeDataVizConverter
CollegeDataStatsConverter

capstone/src/main/java/com/google/univiz/api/VisResource.java
```diff
@@ -1,8 +1,8 @@
 package com.google.univiz.api;
```
kaveh096, line 1: Note that you're moving this class in #77. Make sure it ends up in the right place after resolving the merge conflict.

capstone/src/main/java/com/google/univiz/api/dataviz/CollegeStats.java
```diff
@@ -0,0 +1,46 @@
+package com.google.univiz.api.dataviz;
```
kaveh096, line 1: After #77, we won't have dataviz package anymore. Make sure to move this into the right place i.e. api.representation.

capstone/src/main/java/com/google/univiz/api/dataviz/CollegeDataStatsConverter.java
```diff
@@ -0,0 +1,26 @@
+package com.google.univiz.api.dataviz;
```
kaveh096, line 1: Since #77, i'm not sure what is the right place for this class and in general Converter<> classes. I feel like univiz.api.resource is the right place. It'll be next to the `*Impl.java` class that is going to use it.

PR #82, opened 2020-08-13
College data api impl

ihsan314:
Add a `CollegeDataApiImpl` class to take care of REST calls to the CollegeScoreCard API. Addresses the first goal of #26.

~Due to my incompetence with Guice the tests are failing, I hope to uncover what I got wrong but can't seem to do so as of yet.~ Wasn't a Guice problem, it was a Gson problem, and thanks to Chris it got resolved.

Additionally CollegeId is included in this PR to ensure that the PR builds, but once #85 is complete I will fast-forward this branch and adjust the import statements.


capstone/src/main/java/com/google/univiz/scorecard/CollegeDataApiImpl.java
```diff
@@ -0,0 +1,37 @@
...
+    this.scorecardConverter = scorecardConverter;
+    this.readerProvider = readerProvider;
+  }
+
+  @Override
+  public List<CollegeData> getCollegesById(List<CollegeId> ids) {
+    List<CollegeData> colleges = new ArrayList<>();
+    for (CollegeId id : ids) {
+      // Then deserialize into a ScorecardResponse
+      Gson gson = new Gson();
```
ccondit, line 27: I think your issue is that Gson doesn't know about the autovalue types.
```suggestion
      Gson gson = new GsonBuilder().registerTypeAdapterFactory(GenerateTypeAdapter.FACTORY).create();
```

capstone/src/test/java/com/google/univiz/scorecard/CollegeDataApiImplTest.java
```diff
@@ -0,0 +1,56 @@
...
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnit;
+import org.mockito.junit.MockitoRule;
+
+@RunWith(JUnit4.class)
+public final class CollegeDataApiImplTest {
+
+  @Rule public MockitoRule rule = MockitoJUnit.rule();
+  @Bind @Mock CollegeIdReaderProvider mockReaderProvider;
+  @Bind ScorecardConverter scorecardConverter = new ScorecardConverter();
```
ccondit, line 34: I don't think you'll need this. Guice will be able to inject it by default.

capstone/src/test/java/com/google/univiz/scorecard/CollegeDataApiImplTest.java
```diff
@@ -0,0 +1,56 @@
...
+  @Inject CollegeDataApiImpl testImpl;
```
ccondit, line 35: nit - +private

capstone/src/main/java/com/google/univiz/scorecard/CollegeIdReaderProviderImpl.java
```diff
@@ -0,0 +1,61 @@
...
+      // log this
+      // then create empty stream
+      is =
+          new InputStream() {
+            @Override
+            public int read() {
+              return 0;
+            }
+          };
+    } catch (IOException e) {
```
ccondit, line 48: I would throw an AssertionError here. This shouldn't happen.

`throw new AssertionError(e);`

capstone/src/main/java/com/google/univiz/scorecard/CollegeIdReaderProviderImpl.java
```diff
@@ -0,0 +1,61 @@
...
+import java.io.Reader;
+import java.net.MalformedURLException;
+import java.net.URL;
+import javax.inject.Inject;
+
+class CollegeIdReaderProviderImpl implements CollegeIdReaderProvider {
+  private final UnivizConfig univizConfig;
+
+  @Inject
+  public CollegeIdReaderProviderImpl(UnivizConfig univizConfig) {
```
ccondit, line 17: nit - make this constructor package private.

capstone/src/main/java/com/google/univiz/scorecard/CollegeIdReaderProviderImpl.java
```diff
@@ -0,0 +1,61 @@
...
+    this.univizConfig = univizConfig;
+  }
+
+  @Override
+  public Reader getReaderFromCollegeId(CollegeId id) {
```
ccondit, line 22: FWIW - I think this class should be lightweight as possible - including the URL generation (that way we can verify that the URL is generated correctly in test):

interface CollegeIdReaderProvider {
  InputStream getStreamFromUrl(String URL) throws IOException;
}

Plus if we use a strategy like this we can mock the exceptional cases...

capstone/src/main/java/com/google/univiz/scorecard/CollegeDataApiImpl.java
```diff
@@ -0,0 +1,39 @@
...
+import com.ryanharter.auto.value.gson.GenerateTypeAdapter;
+import java.util.ArrayList;
+import java.util.List;
+import javax.inject.Inject;
+
+public class CollegeDataApiImpl implements CollegeDataApi {
+  private final ScorecardConverter scorecardConverter;
+  private final CollegeIdReaderProvider readerProvider;
+
+  @Inject
```
ccondit, line 17: Nice!!

capstone/src/main/java/com/google/univiz/scorecard/CollegeDataApiImpl.java
```diff
@@ -0,0 +1,39 @@
...
+  public CollegeDataApiImpl(
```
ccondit, line 18: nit - not public on the constructor

capstone/src/main/java/com/google/univiz/scorecard/CollegeDataApiImpl.java
```diff
@@ -0,0 +1,39 @@
...
+import com.google.gson.GsonBuilder;
+import com.google.univiz.CollegeData;
+import com.google.univiz.CollegeDataApi;
+import com.google.univiz.api.CollegeId;
+import com.ryanharter.auto.value.gson.GenerateTypeAdapter;
+import java.util.ArrayList;
+import java.util.List;
+import javax.inject.Inject;
+
+public class CollegeDataApiImpl implements CollegeDataApi {
```
ccondit, line 13: nit - s/public/final/

capstone/src/main/java/com/google/univiz/scorecard/CollegeDataApiImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    this.scorecardConverter = scorecardConverter;
+    this.readerProvider = readerProvider;
+  }
+
+  @Override
+  public List<CollegeData> getCollegesById(List<CollegeId> ids) {
+    List<CollegeData> colleges = new ArrayList<>();
+    for (CollegeId id : ids) {
+      // Then deserialize into a ScorecardResponse
+      Gson gson = new GsonBuilder().registerTypeAdapterFactory(GenerateTypeAdapter.FACTORY).create();
```
ccondit: Move this into a private final instance member. Then we can discuss injection later...

capstone/src/main/java/com/google/univiz/scorecard/CollegeDataApiImpl.java
```diff
@@ -0,0 +1,39 @@
...
+  public CollegeDataApiImpl(
+      ScorecardConverter scorecardConverter, CollegeIdReaderProvider readerProvider) {
+    this.scorecardConverter = scorecardConverter;
+    this.readerProvider = readerProvider;
+  }
+
+  @Override
+  public List<CollegeData> getCollegesById(List<CollegeId> ids) {
+    List<CollegeData> colleges = new ArrayList<>();
+    for (CollegeId id : ids) {
```
ccondit, line 27: Philosophical question - they way this is configured we're going to end up sending a REST request for every college id in serial. This might slow things down a bit. 

Can you think of a way to achieve the same result but with a single request?

capstone/src/main/java/com/google/univiz/scorecard/CollegeDataApiImpl.java
```diff
@@ -0,0 +1,39 @@
...
+      // Then deserialize into a ScorecardResponse
+      Gson gson = new GsonBuilder().registerTypeAdapterFactory(GenerateTypeAdapter.FACTORY).create();
+      ScorecardResponse scorecardResponse =
+          gson.fromJson(readerProvider.getReaderFromCollegeId(id), ScorecardResponse.class);
+      ScorecardData scorecard = scorecardResponse.scorecardData().get(0);
+      // Then convert to a CollegeData
+      CollegeData college = scorecardConverter.convert(scorecard);
```
ccondit, line 35: This seems like a great candidate for some streams - but we can chat about that when we get there...

capstone/src/main/java/com/google/univiz/scorecard/CollegeIdReaderProviderImpl.java
```diff
@@ -0,0 +1,61 @@
...
+    } catch (IOException e) {
+      // log this
+      // then create empty stream
+      is =
+          new InputStream() {
+            @Override
+            public int read() {
+              return 0;
+            }
+          };
```
ccondit, line 57: just throw this back to the caller - we can test and handle it there...

capstone/src/test/java/com/google/univiz/scorecard/CollegeDataApiImplTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+  @Before
+  public void setup() {
+    Guice.createInjector(BoundFieldModule.of(this)).injectMembers(this);
+  }
+
+  @Test
```
ccondit, line 42: How about some other test cases:
* Empty results?
* Multiple results?
* IOExceptions?

capstone/src/test/java/com/google/univiz/scorecard/CollegeDataApiImplTest.java
```diff
@@ -0,0 +1,56 @@
...
+  public void testGetCollegesById() throws IOException {
+    Gson gson = new GsonBuilder().registerTypeAdapterFactory(GenerateTypeAdapter.FACTORY).create();
+    InputStreamReader scorecardReader =
+        new InputStreamReader(
+            Resources.getResource(CollegeDataApiImplTest.class, "scorecard_response.json")
+                .openStream());
+    CollegeId collegeId = CollegeId.create(193900);
```
ccondit, line 49: As you add more tests I think you'll find that moving some of these variables into constants or class members will make things more readable :)

capstone/src/main/java/com/google/univiz/scorecard/CollegeDataApiImpl.java
ihsan314: Ah that's a good point, I think I can just build up the url to include all the ids, and then make one REST call, then do deserialization/conversion/the like. I shall do that instead of this.

capstone/src/main/java/com/google/univiz/scorecard/CollegeDataApiImpl.java
ihsan314: Yea I was thinking about using streams for this, but I'll adjust the loop to do only one REST call first and address the other things.

capstone/src/test/java/com/google/univiz/scorecard/CollegeDataApiImplTest.java
ihsan314: Ah yes, so far I only catered towards the one test I have.

capstone/src/test/java/com/google/univiz/scorecard/CollegeDataApiImplTest.java
ihsan314: They're coming soon.

capstone/src/main/java/com/google/univiz/scorecard/CollegeIdReaderProviderImpl.java
ihsan314: Ah yes, that will be useful, will do.

capstone/src/main/java/com/google/univiz/scorecard/CollegeIdReaderProviderImpl.java
```diff
@@ -0,0 +1,61 @@
...
+  public Reader getReaderFromCollegeId(CollegeId id) {
+    String urlString = "https://api.data.gov/ed/collegescorecard/v1/schools.json?id=";
+    urlString += Integer.toString(id.id());
+    urlString += "&per_page=1&fields=id,school.name,school.city,school.main_";
+    urlString += "campus,location.lat,location.lon,school.carnegie_size_setting,";
+    urlString += "latest.admissions.admission_rate.overall,";
+    urlString += "latest.admissions.sat_scores.average.overall,latest.student.size,";
+    urlString += "latest.cost.attendance.academic_year,latest.student.demographics.men,";
+    urlString += "latest.student.demographics.women&api_key=";
+    urlString += univizConfig.scorecardApiKey();
```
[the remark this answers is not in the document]
ihsan314: That would be less error prone. I am not currently aware of how to do that, but I'll think about it after this PR works out.

capstone/src/test/java/com/google/univiz/scorecard/CollegeDataApiImplTest.java
ihsan314: After experimentation, this didn't work.

ihsan314: Am I missing something in the `setup` function?

capstone/src/main/java/com/google/univiz/scorecard/CollegeIdReaderProviderImpl.java
ihsan314: I threw an `AssertionError` in the case of a `MalformedURLException`, and let the `IOException` be handled elsewhere.

capstone/src/main/java/com/google/univiz/scorecard/CollegeDataApiImpl.java
```diff
@@ -0,0 +1,40 @@
...
+  @Override
+  public List<CollegeData> getCollegesById(List<CollegeId> ids) {
+    List<CollegeData> colleges = new ArrayList<>();
+    for (CollegeId id : ids) {
+      // Then deserialize into a ScorecardResponse
+      Gson gson =
+          new GsonBuilder().registerTypeAdapterFactory(GenerateTypeAdapter.FACTORY).create();
+      ScorecardResponse scorecardResponse =
+          gson.fromJson(readerProvider.getReaderFromCollegeId(id), ScorecardResponse.class);
+      ScorecardData scorecard = scorecardResponse.scorecardData().get(0);
```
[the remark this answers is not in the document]
ihsan314: I have the assertion that the size is one, wouldn't that fail the test automatically?

ihsan314:
So far I think I addressed all concerns besides additional test cases. If everything else looks good, I'll proceed and write more tests.

capstone/src/test/java/com/google/univiz/scorecard/CollegeDataApiImplTest.java
```diff
@@ -0,0 +1,53 @@
...
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnit;
+import org.mockito.junit.MockitoRule;
+
+@RunWith(JUnit4.class)
+public final class CollegeDataApiImplTest {
+
+  @Rule public final MockitoRule rule = MockitoJUnit.rule();
+  @Bind @Mock private CollegeIdReaderProvider mockReaderProvider;
+  @Bind private ScorecardConverter scorecardConverter = new ScorecardConverter();
```
ccondit, line 31: nit - you can remove this because Guice can bind it automatically (via the no arg constructor).

capstone/src/test/java/com/google/univiz/scorecard/CollegeDataApiImplTest.java
ccondit: In a follow up PR? I'm fine for you to add them in this PR :)

capstone/src/main/java/com/google/univiz/scorecard/CollegeDataApiImpl.java
```diff
@@ -0,0 +1,38 @@
...
+      ScorecardConverter scorecardConverter, CollegeIdReaderProvider readerProvider) {
+    this.scorecardConverter = scorecardConverter;
+    this.readerProvider = readerProvider;
+    this.gson = new GsonBuilder().registerTypeAdapterFactory(GenerateTypeAdapter.FACTORY).create();
+  }
+
+  @Override
+  public List<CollegeData> getCollegesById(List<CollegeId> ids) throws IOException {
+    List<CollegeData> colleges = new ArrayList<>();
+    // Then deserialize into a ScorecardResponse
```
ccondit, line 30: nit - i think you can safely drop this comment. the code is very clear :)

capstone/src/main/java/com/google/univiz/scorecard/CollegeDataApiImpl.java
```diff
@@ -0,0 +1,38 @@
...
+    ScorecardResponse scorecardResponse =
+        gson.fromJson(readerProvider.getReaderFromCollegeIds(ids), ScorecardResponse.class);
+    scorecardResponse.scorecardData().stream()
+        .map(scorecard -> scorecardConverter.convert(scorecard))
```
ccondit, line 34: nice - nit use a method reference here:
```suggestion
        .map(scorecardConverter::convert)
```

capstone/src/main/java/com/google/univiz/scorecard/CollegeDataApiImpl.java
```diff
@@ -0,0 +1,38 @@
...
+        .forEach(college -> colleges.add(college));
```
ccondit, line 35: nit - instead of using `forEach` you could just return the collected list:
```suggestion
        .collect(toList());
```

capstone/src/test/java/com/google/univiz/scorecard/CollegeDataApiImplTest.java
ccondit: This works for me when I tried it locally. What's the error that you're observing?

capstone/src/test/java/com/google/univiz/scorecard/CollegeDataApiImplTest.java
ihsan314: Yea in this PR I'll add them.

capstone/src/main/java/com/google/univiz/scorecard/CollegeIdReaderProviderImpl.java
```diff
@@ -0,0 +1,49 @@
...
+    this.univizConfig = univizConfig;
+  }
+
+  @Override
+  public Reader getReaderFromCollegeIds(List<CollegeId> ids) throws IOException {
+    String urlString = "https://api.data.gov/ed/collegescorecard/v1/schools.json?id=";
+    for (CollegeId id : ids) {
+      urlString += Integer.toString(id.id());
+      urlString += ",";
+    }
```
ccondit, line 28: Have a look at [StringJoiner](https://docs.oracle.com/javase/8/docs/api/java/util/StringJoiner.html) to simplify this logic.

capstone/src/main/java/com/google/univiz/scorecard/CollegeIdReaderProviderImpl.java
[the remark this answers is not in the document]
ccondit: In this PR it might be nice to break them out into a Set and use StringJoiner to assemble the `fields` list for simplification...

capstone/src/main/java/com/google/univiz/scorecard/CollegeIdReaderProviderImpl.java
```diff
@@ -0,0 +1,49 @@
...
+  private final UnivizConfig univizConfig;
+
+  @Inject
+  protected CollegeIdReaderProviderImpl(UnivizConfig univizConfig) {
+    this.univizConfig = univizConfig;
+  }
+
+  @Override
+  public Reader getReaderFromCollegeIds(List<CollegeId> ids) throws IOException {
+    String urlString = "https://api.data.gov/ed/collegescorecard/v1/schools.json?id=";
```
ccondit, line 24: Consider making this a constant.

capstone/src/test/java/com/google/univiz/scorecard/CollegeDataApiImplTest.java
ihsan314:
```
1) Binding to null values is only allowed for fields that are annotated @Nullable.
  at com.google.univiz.scorecard.CollegeDataApiImplTest.scorecardConverter(CollegeDataApiImplTest.java:27)
```

capstone/src/main/java/com/google/univiz/scorecard/CollegeIdReaderProviderImpl.java
```diff
@@ -0,0 +1,49 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.Reader;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.util.List;
+import javax.inject.Inject;
+
+class CollegeIdReaderProviderImpl implements CollegeIdReaderProvider {
```
ccondit, line 14: This seam that you created for testing is awesome - but it still has a lot of untested magic in it:
* Making sure that the constructed URL string is correct
  - has the right API key, and IDs
* that an inputstreamreader is wrapping the result (although I'm less worried about this)

a couple of options:
- Change the return type of the seam to InputStream
- Change the input to the provider to be a String - then you can verify that the String is computed correctly in the client of this provider

Also - can you think of a way to unit test this implementation (without connecting to a remote REST resource)?

Feel free to chat with me offline about this comment.

ccondit: @mmurray22 - this seam will be useful for #83, as well. Perhaps you can share this as common infrastructure?

kaveh096: +1 You can probably just move getReaderFromCollegeIds back to CollegeDataApiImpl so it can be tested.

capstone/src/main/java/com/google/univiz/scorecard/CollegeIdReaderProviderImpl.java
```diff
@@ -0,0 +1,66 @@
...
+    // Make REST call
+    return new InputStreamReader(getStreamFromUrl(urlStringBuilder.toString()));
+  }
+
+  public InputStream getStreamFromUrl(String urlString) throws IOException {
+    InputStream is;
+    try {
+      URL url = new URL(urlString);
+      is = url.openStream();
+      return is;
```
kaveh096, line 61: nit: inline

```suggestion
      return url.openStream();
```

capstone/src/main/java/com/google/univiz/scorecard/CollegeIdReaderProviderImpl.java
```diff
@@ -0,0 +1,66 @@
...
+    fields.add("school.main_campus");
+    fields.add("location.lat");
+    fields.add("location.lon");
+    fields.add("school.carnegie_size_setting");
+    fields.add("latest.admissions.admission_rate_overall");
+    fields.add("latest.admissions.sat_scores.average.overall");
+    fields.add("latest.student.size");
+    fields.add("latest.cost.attendance.academic_year");
+    fields.add("latest.student.demographics.men");
+    fields.add("latest.student.demographics.women");
```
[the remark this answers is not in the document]
ihsan314: I think this addresses [your earlier question](https://github.com/googleinterns/univiz/pull/82#discussion_r470915661) then. Yea this would be particularly useful. I'll integrate this in.

capstone/src/test/java/com/google/univiz/scorecard/CollegeDataApiImplTest.java
ccondit: Did these get added?

ihsan314: These are still pending.

capstone/src/test/java/com/google/univiz/scorecard/URLProviderImplTest.java
```diff
@@ -0,0 +1,52 @@
...
+  public void setup() {
+    Guice.createInjector(new UnivizConfigModule(), BoundFieldModule.of(this)).injectMembers(this);
+  }
+
+  @Test
+  public void testGetUrlFromCollegeIds() {
+    CollegeId fakeCollegeId1 = CollegeId.create(1);
+    CollegeId fakeCollegeId2 = CollegeId.create(2);
+    CollegeId fakeCollegeId3 = CollegeId.create(3);
+    List<CollegeId> ids = Arrays.asList(fakeCollegeId1, fakeCollegeId2, fakeCollegeId3);
```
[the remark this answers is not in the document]
ihsan314: Will check it out.

capstone/src/test/java/com/google/univiz/scorecard/URLProviderImplTest.java
```diff
@@ -0,0 +1,52 @@
...
+        new StringBuilder("https://api.data.gov/ed/collegescorecard/v1/schools.json?id=");
+    expectedBuilder.append(
+        String.format("%d,%d,%d", fakeCollegeId1.id(), fakeCollegeId2.id(), fakeCollegeId3.id()));
+    expectedBuilder.append("&per_page=1&fields=");
+    expectedBuilder.append("id,school.name,school.city,school.main_campus,");
+    expectedBuilder.append(
+        "location.lat,location.lon,school.carnegie_size_setting,latest.admissions.admission_rate.overall,");
+    expectedBuilder.append("latest.admissions.sat_scores.average.overall,latest.student.size,");
+    expectedBuilder.append("latest.cost.attendance.academic_year,latest.student.demographics.men,");
+    expectedBuilder.append("latest.student.demographics.women&api_key=");
```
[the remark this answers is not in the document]
ihsan314: I can use a similar technique to `UnivizConfigModuleTest` to verify that the api_key injection is functional.

capstone/src/test/java/com/google/univiz/scorecard/URLProviderImplTest.java
```diff
@@ -0,0 +1,52 @@
...
+    String expected = expectedBuilder.toString();
+
+    assertThat(actual.substring(0, expected.length())).isEqualTo(expected);
```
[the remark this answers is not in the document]
ihsan314: Yea it should help I believe.

capstone/src/main/java/com/google/univiz/scorecard/URLProviderImpl.java
```diff
@@ -0,0 +1,43 @@
...
+            .map(field -> field.getAnnotation(SerializedName.class))
+            .filter(Objects::nonNull)
+            .map(SerializedName::value)
+            .collect(toSet());
+  }
+
+  @Override
+  public String getUrlFromCollegeIds(List<CollegeId> ids) {
+    String stringWithFields = fields.stream().collect(joining(","));
+    final StringBuilder urlStringBuilder = new StringBuilder();
```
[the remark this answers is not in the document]
ihsan314: Will do.

capstone/src/main/java/com/google/univiz/scorecard/URLProviderImpl.java
```diff
@@ -0,0 +1,43 @@
...
+
+  @Inject
+  protected URLProviderImpl(UnivizConfig univizConfig) {
+    this.univizConfig = univizConfig;
+    fields =
+        Arrays.stream(ScorecardData.class.getDeclaredMethods())
+            .map(field -> field.getAnnotation(SerializedName.class))
+            .filter(Objects::nonNull)
+            .map(SerializedName::value)
+            .collect(toSet());
```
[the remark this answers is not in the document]
ihsan314: Ok I'll do that, I think that's why the test is failing.

capstone/src/main/java/com/google/univiz/scorecard/URLProviderImpl.java
```diff
@@ -0,0 +1,43 @@
...
+import com.google.gson.annotations.SerializedName;
+import com.google.univiz.api.CollegeId;
+import com.google.univiz.config.UnivizConfig;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Objects;
+import java.util.Set;
+import javax.inject.Inject;
+
+class URLProviderImpl implements URLProvider {
```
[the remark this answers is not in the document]
ihsan314: Will do.

biancamacias:
this PR is coming along so great, ihsan! 🥇

capstone/src/test/java/com/google/univiz/scorecard/CollegeDataApiImplTest.java
ihsan314: I've added the first two, and am currently figuring out how to add the IOExceptions one.

capstone/src/main/java/com/google/univiz/scorecard/CollegeIdReaderProvider.java
```diff
@@ -0,0 +1,9 @@
+package com.google.univiz.scorecard;
+
+import java.io.IOException;
+import java.io.InputStream;
+
+interface CollegeIdReaderProvider {
+
+  InputStream getStreamFromUrl(String urlString) throws IOException;
```
kaveh096, line 8: nit: javadoc for interface and method.

capstone/src/main/java/com/google/univiz/scorecard/URLProvider.java
```diff
@@ -0,0 +1,9 @@
+package com.google.univiz.scorecard;
+
+import com.google.univiz.api.CollegeId;
+import java.util.List;
+
+interface URLProvider {
+
+  String getUrlFromCollegeIds(List<CollegeId> ids);
```
kaveh096, line 8: ditto javadoc

capstone/src/main/java/com/google/univiz/scorecard/URLProviderImpl.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.Objects;
+import java.util.Set;
+import javax.inject.Inject;
+
+final class URLProviderImpl implements URLProvider {
+  private final UnivizConfig univizConfig;
+  private final Set<String> fields;
+
+  @Inject
+  protected URLProviderImpl(UnivizConfig univizConfig) {
```
kaveh096, line 21: I don't think you need to make it protected. Protected means by sub-classes and package classes. Since this is a final class, it cannot have sub-classes. So package-private is all you need.
```suggestion
  URLProviderImpl(UnivizConfig univizConfig) {
```

capstone/src/main/java/com/google/univiz/scorecard/URLProviderImpl.java
```diff
@@ -0,0 +1,45 @@
...
+    this.univizConfig = univizConfig;
+    fields =
+        Arrays.stream(ScorecardData.class.getDeclaredMethods())
+            .map(field -> field.getAnnotation(SerializedName.class))
+            .filter(Objects::nonNull)
+            .map(SerializedName::value)
```
kaveh096, line 27: If you're still worried about non-determinism, you can add a `.sorted()` right before collect.

capstone/src/main/java/com/google/univiz/scorecard/URLProviderImpl.java
```diff
@@ -0,0 +1,45 @@
...
+
+  @Override
+  public String getUrlFromCollegeIds(List<CollegeId> ids) {
+    String stringWithFields = fields.stream().collect(joining(","));
+    StringBuilder urlStringBuilder = new StringBuilder();
+    urlStringBuilder.append("https://api.data.gov/ed/collegescorecard/v1/schools.json?id=");
+    String stringWithIds =
+        ids.stream().map(CollegeId::id).map(String::valueOf).collect(joining(","));
+    urlStringBuilder.append(stringWithIds);
+    urlStringBuilder.append("&per_page=1&fields=");
```
kaveh096, line 39: Should the `per_page` be a larger number like `ids.size()`? We want to make sure we get all colleges on the first page, otherwise we have to make subsequent calls to get the 2nd, 3rd pages.

capstone/src/test/java/com/google/univiz/scorecard/CollegeDataApiImplTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+  @Test
+  public void testGetCollegesById() throws IOException {
+    CollegeId collegeId = CollegeId.create(193900);
+    String scorecardUrlString =
+        Resources.getResource(CollegeDataApiImplTest.class, "scorecard_response.json").toString();
+    when(mockUrlProvider.getUrlFromCollegeIds(Arrays.asList(collegeId)))
+        .thenReturn(scorecardUrlString);
+    when(mockReaderProvider.getStreamFromUrl(scorecardUrlString))
+        .then(invocation -> readerProvider.getStreamFromUrl(scorecardUrlString));
```
ccondit, line 49: This is a very interesting approach! I expected you to mock the readerProvider and leave url provider as is. Through your use of mocks delegating to an actual instance, you've effectively discovered [spies](https://stackoverflow.com/questions/15052984/what-is-the-difference-between-mocking-and-spying-when-using-mockito).

Here's what I would do to simplify things:

- Delete your mock of `CollegeIdReaderProvider`
- Switch to: `@Bind @Spy private CollegeIdReaderProvider readerProvider = new CollegeIdReaderProviderImpl();`
- Make `CollegeIdReaderProviderImpl` non final s.t. we can spy on it.
- Stop mocking `readerProvider` in all tests except for the IO test.
- Switch the mockito setup in the IOException test to mock out the `readerProvider`.

capstone/src/main/java/com/google/univiz/scorecard/URLProviderImpl.java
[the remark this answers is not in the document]
ccondit: heh - kaveh beat me to it :)

capstone/src/main/java/com/google/univiz/scorecard/URLProviderImpl.java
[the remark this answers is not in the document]
ccondit: heh - kaveh beat me to it :)

capstone/src/test/java/com/google/univiz/scorecard/CollegeDataApiImplTest.java
ihsan314: Ah ok this makes it simpler.

capstone/src/main/java/com/google/univiz/scorecard/URLProviderImpl.java
ihsan314: Yea I didn't consider that initially, I'll add that in now.

capstone/src/test/java/com/google/univiz/scorecard/CollegeDataApiImplTest.java
```diff
@@ -0,0 +1,94 @@
...
+    when(mockUrlProvider.getUrlFromCollegeIds(Arrays.asList(collegeId)))
+        .thenReturn(scorecardUrlString);
+    List<CollegeData> colleges = testImpl.getCollegesById(Arrays.asList(collegeId));
+    assertThat(colleges).isEmpty();
+  }
+
+  @Test
+  public void testMultipleResults() throws IOException {
+    CollegeId collegeId = CollegeId.create(193900);
+    CollegeId collegeId2 = CollegeId.create(13900);
```
ccondit, line 69: nit - collegeId2 is never used :)

capstone/src/test/java/com/google/univiz/scorecard/URLProviderImplTest.java
```diff
@@ -0,0 +1,68 @@
...
+    expectedFields.add("latest.cost.attendance.academic_year");
+    expectedFields.add("latest.student.demographics.men");
+    expectedFields.add("latest.student.demographics.women");
+
+    StringBuilder expectedBuilder =
+        new StringBuilder("https://api.data.gov/ed/collegescorecard/v1/schools.json?=");
+    expectedBuilder.append(
+        String.format("%d,%d,%d", fakeCollegeId1.id(), fakeCollegeId2.id(), fakeCollegeId3.id()));
+    expectedBuilder.append("&per_page=3&fields=");
+    expectedBuilder.append(expectedFields.stream().map(field -> "").collect(joining(",")));
```
ccondit, line 60: Can this be stronger now? Seems like we shouldn't ignore the field names now that the ordering is deterministic.

ihsan314: Ah yea I'll make the test stronger.

capstone/src/test/java/com/google/univiz/scorecard/CollegeDataApiImplTest.java
ihsan314: Oh I didn't notice that at first, I'll remove it.

capstone/src/main/java/com/google/univiz/api/CollegeId.java
```diff
@@ -1,8 +1,18 @@
 package com.google.univiz.api;
```
kaveh096, line 1: Don't forget to move this to `univiz.api.representation` package.

capstone/src/main/java/com/google/univiz/api/MapsResource.java
```diff
@@ -4,8 +4,8 @@
 import java.util.List;
 
 /**
- * The MapsResource interface is responsible for all data management/processing specific for the maps
- * feature of the UniViz webapp.
+ * The MapsResource interface is responsible for all data management/processing specific for the
+ * maps feature of the UniViz webapp.
```
kaveh096, line 8: It looks like your only change to this file is javadoc reformatting. Considering that it has now moved to the `api.resource` package you can revert your reformatting change to avoid extra conflict resolution. Or just move the class in our local branch has well.

capstone/src/main/java/com/google/univiz/scorecard/CollegeDataApiImpl.java
```diff
@@ -0,0 +1,42 @@
...
+import javax.inject.Inject;
+
+final class CollegeDataApiImpl implements CollegeDataApi {
+  private final ScorecardConverter scorecardConverter;
+  private final CollegeIdReaderProvider readerProvider;
+  private final URLProvider urlProvider;
+  private final Gson gson;
+
+  @Inject
+  protected CollegeDataApiImpl(
```
kaveh096, line 21: No need to make it protected. The class is final so it cannot have sub-classes.
```suggestion
  CollegeDataApiImpl(
```

capstone/src/main/java/com/google/univiz/scorecard/CollegeIdReaderProviderImpl.java
```diff
@@ -0,0 +1,17 @@
+package com.google.univiz.scorecard;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.MalformedURLException;
+import java.net.URL;
+
+class CollegeIdReaderProviderImpl implements CollegeIdReaderProvider {
+  public InputStream getStreamFromUrl(String urlString) throws IOException {
```
kaveh096, line 9:
```suggestion
  @Override
  public InputStream getStreamFromUrl(String urlString) throws IOException {
```

kaveh096:
LGTM + few more comments.

PR #83, opened 2020-08-13
Implementation of Suggestion API Interface

mmurray22:
Initial implementation of Suggestion API Interface. Minimal Javadoc provided, more Javadoc can be added. No complex exception handling will be addressed in this PR. Compiles. This PR makes advancements on the groundwork laid out in https://github.com/googleinterns/univiz/pull/82


capstone/src/main/java/com/google/univiz/SuggestionApiReaderProvider.java
```diff
@@ -0,0 +1,8 @@
+package com.google.univiz;
+
+import java.io.IOException;
+import java.io.InputStream;
+
+interface SuggestionApiReaderProvider {
+  InputStream getStreamFromUrl(String urlString) throws IOException;
```
ihsan314, line 7: This interface and it's implementation are doing the same thing as `CollegeIdReaderProvider` and `CollegeIdReaderProviderImpl` in #82. Unnecessary duplication is probably bad, so we should stick with either your interface or mine.

capstone/src/main/java/com/google/univiz/SuggestionDataApiImpl.java
```diff
@@ -0,0 +1,34 @@
...
+
+  @Inject
+  protected SuggestionDataApiImpl(
+      UrlProvider urlProvider, SuggestionApiReaderProvider readerProvider, Gson gson) {
+    this.urlProvider = urlProvider;
+    this.readerProvider = readerProvider;
+    this.gson = gson;
+  }
+
+  /** Takes REST API Json response and converts it to SuggestionData */
```
biancamacias, line 22:
```suggestion
  // Takes REST API Json response and converts it to SuggestionData
```
i think `//` is meant for code comments while `**/*/` implies javadoc.

capstone/src/main/java/com/google/univiz/api/SuggestionDataApiImpl.java
```diff
@@ -0,0 +1,35 @@
+package com.google.univiz.api;
```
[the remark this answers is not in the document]
mmurray22: done!

capstone/src/test/resources/com/google/univiz/scorecard/suggestion-response.json
```diff
@@ -0,0 +1,15 @@
+{
```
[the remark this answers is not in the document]
mmurray22: This file is relevant for a different test-I think it's supposed to be in a different folder

capstone/src/main/java/com/google/univiz/scorecard/CollegeDataApiImpl.java
```diff
@@ -32,8 +32,7 @@
   @Override
   public List<CollegeData> getCollegesById(List<CollegeId> ids) throws IOException {
     InputStreamReader reader =
-        new InputStreamReader(
-            readerProvider.getStreamFromUrl(urlProvider.getUrlFromCollegeIds(ids)));
+        new InputStreamReader(readerProvider.getStreamFromUrl(urlProvider.getUrl(null, ids)));
```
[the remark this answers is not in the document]
mmurray22: done!

capstone/src/test/resources/com/google/univiz/scorecard/suggestion-response.json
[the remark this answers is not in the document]
mmurray22: removed!

capstone/src/test/java/com/google/univiz/scorecard/URLProviderImplTest.java
```diff
@@ -21,6 +21,11 @@
 
 @RunWith(JUnit4.class)
 public final class URLProviderImplTest {
+  private final String frontUrl = "https://api.data.gov/ed/collegescorecard/v1/schools.json?";
+  private final String querySchoolName = "school.name=";
+  private final String queryTypeId = "id=";
+  private final String perPage = "&per_page=";
+  private final String suggestionFields = "&fields=id,school.name";
```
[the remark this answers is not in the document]
mmurray22: updated!

capstone/src/main/java/com/google/univiz/scorecard/URLProviderImpl.java
```diff
@@ -30,17 +36,40 @@
...
+    }
+    urlStringBuilder.append("&api_key=");
+    urlStringBuilder.append(univizConfig.scorecardApiKey());
+    return urlStringBuilder.toString();
+  }
+
+  private StringBuilder buildSuggestionUrl(
+      StringBuilder urlStringBuilder, String partialCollegeName) {
+    urlStringBuilder.append(queryTypeName);
+    urlStringBuilder.append(partialCollegeName);
```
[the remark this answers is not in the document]
mmurray22: added per page :)

capstone/src/main/java/com/google/univiz/scorecard/URLProviderImpl.java
```diff
@@ -14,6 +14,14 @@
 import javax.inject.Inject;
 
 final class URLProviderImpl implements URLProvider {
+  private static final String FRONT_URL =
+      "https://api.data.gov/ed/collegescorecard/v1/schools.json?";
+  private static final String QUERY_TYPE_NAME = "school.name=";
+  private static final String QUERY_TYPE_ID = "id=";
+  private static final String PER_PAGE = "&per_page=";
+  private static final String FIELDS_PARAM = "&fields=";
+  private static final String SUGGESTION_FIELDS = "id,school.name";
```
ccondit, line 23: By hard coding `SUGGESTION_FIELDS` we lose the generic nature of the reflection in the constructor. Can you think of a way to refactor this s.t. we can maintain the ability to inspect the fields from the representation classes in a reusable way?

mmurray22: Hi Chris, thanks for your feedback. I'm a bit confused as to what you mean by refactoring-as of right now, these are the only fields that we want to get back from the colleges. What inspection are you thinking of?

ccondit: Right now, the fields to query when making the REST request are only based on the Gson annotations in `ScorecardData`. So if new fields need to be added, only `ScorecardData` would need to be updated.

If we hard code these fields for `SuggestionData` in this `SUGGESTION_FIELDS` constant, now the fields are replicated in two places.

Is there a way to prevent that? Can the logic in the constructor be used for both? Or even better - can it be used for a generic GSON representation?

mmurray22: done :)

capstone/src/main/java/com/google/univiz/scorecard/SuggestionDataApiImpl.java
```diff
@@ -1,12 +1,34 @@
...
+      URLProvider urlProvider, CollegeIdReaderProvider readerProvider, Gson gson) {
+    this.urlProvider = urlProvider;
+    this.readerProvider = readerProvider;
+    this.gson = gson;
+  }
+
+  /** Takes REST API Json response and converts it to SuggestionData */
+  private SuggestionResponse convertJsonToSuggestionResponse(InputStreamReader suggestionReader) {
+    return gson.fromJson(suggestionReader, SuggestionResponse.class);
+  }
```
kaveh096, line 25: Per [the style guide](https://google.github.io/styleguide/javaguide.html#s3.4.2-ordering-class-contents), class members should follow a logical and consistent ordering. Inside google3 we have more strict and clear guideline but the public style guide is not super clear. So this is just a soft suggestion.

Here are few rules you can use for method ordering:
 + Keep constructors at the top. Starting with the default one. Private constructors are exempt.
 + Followed by public methods, then protected, and finally private.
 + Within each visibility group keep static and non-statics in their own sub-group as much as possible.
 + A caller method should usually be above the callee.
 + Keep overloads always together.
 + Keep methods that are working together to accomplish a task closer to each other.

These rules really help the next person who is reading the class from top to bottom.

capstone/src/main/java/com/google/univiz/scorecard/URLProviderImpl.java
```diff
@@ -14,8 +15,16 @@
...
 final class URLProviderImpl implements URLProvider {
+  private static final String FRONT_URL =
+      "https://api.data.gov/ed/collegescorecard/v1/schools.json?";
+  private static final String QUERY_TYPE_NAME = "school.name=";
+  private static final String QUERY_TYPE_ID = "id=";
+  private static final String PER_PAGE = "&per_page=";
+  private static final String FIELDS_PARAM = "&fields=";
+  private static final String SUGGESTION_PER_PAGE = "20";
   private final UnivizConfig univizConfig;
   private final Set<String> fields;
```
kaveh096, line 26: This variable name is now ambiguous. Rename to something more specific. Like resourceDataFields, dataFields, resourceFields, etc.

capstone/src/main/java/com/google/univiz/scorecard/URLProviderImpl.java
```diff
@@ -14,8 +15,16 @@
...
+  private final Set<String> fieldsSugg;
```
kaveh096, line 27: Avoid acronyms in naming unless it's making the code more readable in some other way. Here there is no reason to use an acronym. Just rename to suggestionFields,

capstone/src/main/java/com/google/univiz/scorecard/URLProviderImpl.java
```diff
@@ -27,17 +36,43 @@
             .map(SerializedName::value)
             .sorted()
             .collect(toCollection(LinkedHashSet::new));
+    fieldsSugg =
+        Arrays.stream(SuggestionData.class.getDeclaredMethods())
+            .map(field -> field.getAnnotation(SerializedName.class))
+            .filter(Objects::nonNull)
+            .map(SerializedName::value)
+            .sorted()
+            .collect(toCollection(LinkedHashSet::new));
```
kaveh096, line 45: This is now duplicated code. Consider pulling out a helper method:

```java
  private static Set<String> getCollect(Class<?> serilizableClass) {
    return Arrays.stream(serilizableClass.getDeclaredMethods())
        .map(field -> field.getAnnotation(SerializedName.class))
        .filter(Objects::nonNull)
        .map(SerializedName::value)
        .sorted()
        .collect(toCollection(LinkedHashSet::new));
  }
```

capstone/src/main/java/com/google/univiz/scorecard/URLProviderImpl.java
mmurray22: ok thanks!

ccondit, line 27: nit - `fields` and `fieldsSugg` are hard to differentiate. Consider refactoring to `collegeDataFields` and `suggestionFields` for clarity :)

capstone/src/main/java/com/google/univiz/scorecard/URLProviderImpl.java
ccondit: +1

capstone/src/main/java/com/google/univiz/scorecard/URLProviderImpl.java
mmurray22: changed it to collegeDataFields

capstone/src/main/java/com/google/univiz/scorecard/URLProviderImpl.java
mmurray22: done!

capstone/src/main/java/com/google/univiz/scorecard/URLProviderImpl.java
mmurray22: done!

PR #84, opened 2020-08-13
SuggestionData implementation

mmurray22:
What is SuggestionData?
SuggestionData is the object the data from the college data source (currently Scorecard API) will be translated into after being received as a JSONObject.

Overview of this implementation?
Uses AutoValue and GenerateTypeAdapter to create the SuggestionData object.  

Where will this object be used?
In SuggestionDataApi and SearchResource

Note: Moved SuggestionData to the scorecard package.


capstone/src/main/java/com/google/univiz/SuggestionData.java
```diff
@@ -1,4 +1,18 @@
...
-public class SuggestionData {}
+@GenerateTypeAdapter
+@AutoValue
+abstract class SuggestionData {
+  @SerializedName("school.name")
+  @Nullable
+  abstract String name();
+
+  @SerializedName("id")
+  abstract int id();
```
ihsan314, line 17: I guess this works for now, I doubt any code we have that deserializes into a CollegeId, and such a feature doesn't seem necessary.

ccondit:
>  When I tried using the CollegeId data class for the id field, I was unable to make it compatible with the testing setup due to its current form.

What was the incompatibility?

capstone/src/main/java/com/google/univiz/SuggestionData.java
```diff
@@ -1,4 +1,18 @@
...
+import com.ryanharter.auto.value.gson.GenerateTypeAdapter;
+import javax.annotation.Nullable;
+
 /** SuggestionData represents the college data specific to the Search feature */
-public class SuggestionData {}
+@GenerateTypeAdapter
+@AutoValue
+abstract class SuggestionData {
+  @SerializedName("school.name")
+  @Nullable
```
kaveh096, line 13: Why nullable? Returning null would make it difficult to handle errors when we want to show auto complete data in JS.

mmurray22: Ah ok...what happens if I don't make it nullable-> would that throw an error right there and then? I was thinking I could check earlier up in the pipeline, but I can change that

mmurray22:
> > When I tried using the CollegeId data class for the id field, I was unable to make it compatible with the testing setup due to its current form.
> 
> What was the incompatibility?

When I tried to make the CollegeId the object it converted to, it didn't have the same value as the json that was parsing into the SuggestionData. CollegeId also had a create function and other parts that made it more complex than just using the int at first..however, I will look into how to address this problem and update this thread.

biancamacias:
> > > When I tried using the CollegeId data class for the id field, I was unable to make it compatible with the testing setup due to its current form.
> > 
> > 
> > What was the incompatibility?
> 
> When I tried to make the CollegeId the object it converted to, it didn't have the same value as the json that was parsing into the SuggestionData. CollegeId also had a create function and other parts that made it more complex than just using the int at first..however, I will look into how to address this problem and update this thread.

where are you having this problem? i think i can help you. `CollegeId` does differ from all the other classes because it doesnt use a builder.

PR #86, opened 2020-08-13
SuggestionResponse Implementation

mmurray22:
What is SuggestionResource?
SuggestionResource is the object the data from the college data source (currently Scorecard API) will be translated into after being received as a JSONObject.

Overview of this implementation?
Uses AutoValue and GenerateTypeAdapter to create the SuggestionResource object.

Note: Solved PR overlap problem! 

Where will this object be used?
In SuggestionDataApi and SearchResource


capstone/src/main/java/com/google/univiz/SuggestionData.java
```diff
@@ -1,4 +1,18 @@
...
+import com.google.gson.annotations.SerializedName;
+import com.ryanharter.auto.value.gson.GenerateTypeAdapter;
+import javax.annotation.Nullable;
+
 /** SuggestionData represents the college data specific to the Search feature */
-public class SuggestionData {}
+@GenerateTypeAdapter
+@AutoValue
+abstract class SuggestionData {
+  @SerializedName("school.name")
```
ccondit, line 12: Just noting that there's not a need to provide `@SerliazedName` unless you want to change the [default mapping](https://github.com/rharter/auto-value-gson#usage).

This comment applies to all usages of @SerializedName in this PR.

capstone/src/main/java/com/google/univiz/SuggestionData.java
```diff
@@ -1,4 +1,18 @@
...
+  @Nullable
+  abstract String name();
+
+  @SerializedName("id")
+  abstract int id();
```
ccondit, line 17: Wasn't this going to be CollegeId?

capstone/src/main/java/com/google/univiz/SuggestionResponse.java
```diff
@@ -0,0 +1,14 @@
+package com.google.univiz;
+
+import com.google.auto.value.AutoValue;
+import com.google.gson.annotations.SerializedName;
+import com.ryanharter.auto.value.gson.GenerateTypeAdapter;
+import java.util.List;
+
+/** SuggestionResponse represents the list SuggestionData specific for the Search feature */
```
ccondit, line 8:
```suggestion
/** SuggestionResponse represents the list of SuggestionData specific for the Search feature */
```

capstone/src/test/resources/com/google/univiz/suggestion_response.json
```diff
@@ -0,0 +1,15 @@
+{
+  "metadata":
+  {
+    "total":1,
+    "page":0,
+    "per_page":20
+  },
```
ccondit, line 7: where is this `metadata` coming from?

capstone/src/test/java/com/google/univiz/SuggestionDataTest.java
```diff
@@ -0,0 +1,28 @@
...
+import java.io.InputStreamReader;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public final class SuggestionDataTest {
+  /** Tests whether the JSON properly deserializes into a SuggestionData object */
+  @Test
+  public void testJsonDeserializes() throws Exception {
```
ccondit, line 18: Are we ever going to deserialize this json? Might make more sense to verify that it serializes as expected, since we'll be writing that JSON to the client :)

capstone/src/main/java/com/google/univiz/SuggestionData.java
mmurray22: Yes-I have a note about this in the description!

mmurray22: If you see a way to address the problem in the PR description, we can talk offline and I will incorporate it into the PR

capstone/src/main/java/com/google/univiz/SuggestionResponse.java
```diff
@@ -0,0 +1,14 @@
+package com.google.univiz;
```
kaveh096, line 1: According to the new packaging that @biancamacias is doing, this class should probably go into `api.representation` package. Or it could go into `univiz.scorecard`. Both seems fine to me.

capstone/src/main/java/com/google/univiz/SuggestionResponse.java
```diff
@@ -0,0 +1,14 @@
...
+import com.google.gson.annotations.SerializedName;
+import com.ryanharter.auto.value.gson.GenerateTypeAdapter;
+import java.util.List;
+
+/** SuggestionResponse represents the list SuggestionData specific for the Search feature */
+@GenerateTypeAdapter
+@AutoValue
+abstract class SuggestionResponse {
+  @SerializedName("results")
+  abstract List<SuggestionData> suggestionData();
```
kaveh096, line 13: nit: since it's a list use a plural name. Like `suggestions`.

capstone/src/test/java/com/google/univiz/SuggestionDataTest.java
kaveh096: I think you're confusing SuggestionData with SearchData (as i did before). Or maybe i'm the confused one right now 😂. So this one is the one that we deserialize the scorecard results into. SearchData is the one our servlet returns.

capstone/src/test/resources/com/google/univiz/suggestion_response.json
mmurray22: This metadata section of the JSON is included in all responses from the REST API. I wanted to include it here to make the test as realistic as possible.

capstone/src/main/java/com/google/univiz/SuggestionResponse.java
mmurray22: I think the representation package is probably better, just because this doesn't use nay scorecard functions/structures

capstone/src/test/java/com/google/univiz/SuggestionDataTest.java
mmurray22: Resolved in standup :)

capstone/src/main/java/com/google/univiz/SuggestionResponse.java
mmurray22: haha I will move it to scorecard to be in the same package as suggestiondata 👍

capstone/src/main/java/com/google/univiz/SuggestionData.java
mmurray22: updated! I didn't need the SerializedName for id :)

capstone/src/main/java/com/google/univiz/scorecard/SuggestionResponse.java
```diff
@@ -1,4 +1,14 @@
...
-public class SuggestionResponse {}
+import com.google.auto.value.AutoValue;
+import com.google.gson.annotations.SerializedName;
+import com.ryanharter.auto.value.gson.GenerateTypeAdapter;
+import java.util.List;
+
+/** SuggestionResponse represents the list of SuggestionData specific for the Search feature */
+@GenerateTypeAdapter
+@AutoValue
+public abstract class SuggestionResponse {
```
kaveh096, line 11: does it need to be public?

mmurray22: yes, otherwise I can't use it in the SuggestionDataApi package

PR #88, opened 2020-08-17
Updates to the SuggestionDataApi interface

mmurray22:
Although the interface has been approved already, when I pushed my PRs for SuggestionData, I realized that the best way to handle the List<SuggestionData> return type for the getCollegeSuggestions function is to create the type SuggestionResponse. One reason I think this a better way to handle the return type is because it abstracts out some of the specificity of both the List and the list element type SuggestionData. If for some reason, in the future, we wished not to used SuggestionData and/or wanted to change the List to a different data structure (or even a different type of list), then instead of having to go to a bunch of different locations in the code base and change the List<SuggestionData> return type there, we can just use SuggestionResponse and only have to change it SuggestionResponse. Thus, this PR makes that change to the SuggestionDataApi interface so the following PRs in this chain make more sense/compile


ccondit:
> I realized that the best way to handle the List return type for the getCollegeSuggestions function is to create the type SuggestionResponse.

Can you say more about why this is the "best way"? What problems does returning a `List` present?

mmurray22:
> > I realized that the best way to handle the List return type for the getCollegeSuggestions function is to create the type SuggestionResponse.
> 
> Can you say more about why this is the "best way"? What problems does returning a `List` present?

@ccondit Updated!

capstone/src/main/java/com/google/univiz/SuggestionResponse.java
```diff
@@ -0,0 +1,4 @@
+package com.google.univiz;
+
+/** SuggestionResponse represents the list college data specific to the Search feature */
```
kaveh096, line 3:
```suggestion
/** SuggestionResponse represents the list of college data specific to the Search feature */
```

PR #95, opened 2020-08-19
Search servlet part 1: Add a helper class to handle the gson serialization

kaveh096:


capstone/src/main/java/com/google/univiz/common/ServletHelper.java
```diff
@@ -0,0 +1,18 @@
+package com.google.univiz.common;
+
+import com.google.gson.Gson;
+import java.io.IOException;
+import javax.servlet.http.HttpServletResponse;
+
+/** A set of static utility methods that are common among servlet classes. */
+public final class ServletHelper {
+
+  public static void writeJsonToResponse(Gson gson, HttpServletResponse response, Object results)
```
ccondit, line 10: nit - result
```suggestion
  public static void writeJsonToResponse(Gson gson, HttpServletResponse response, Object result)
```

capstone/src/main/java/com/google/univiz/common/ServletHelper.java
```diff
@@ -0,0 +1,18 @@
...
+      throws IOException {
+    String json = gson.toJson(results);
+    response.setContentType("application/json");
+    response.getWriter().println(json);
```
ccondit, line 14: nit - I *think* we can save a line here with:
`gson.toJson(result, resp.getWriter());`

PR #39, opened 2020-07-31
Add suggested calendar to UniViz

ihsan314:
See #38 for details.

PR #109, opened 2020-08-24
Add Deadline visualization datatype

ihsan314:
Move forward with #39 and #38 by defining the structure of the Deadline datatype, allowing for a heatmap of deadlines for the user's colleges to be added. The deadlines will be hardcoded as the College Scorecard API does not include this information.

### DIFF
--- a/capstone/src/main/java/com/google/univiz/api/representation/Deadline.java
+++ b/capstone/src/main/java/com/google/univiz/api/representation/Deadline.java
@@ -1,6 +1,7 @@
 package com.google.univiz.api.representation;
 
 import com.google.auto.value.AutoValue;
+import com.ryanharter.auto.value.gson.GenerateTypeAdapter;
 
 /**
  * Contains the date when applications are due for a particular college. Will be made serializable
--- a/capstone/src/main/java/com/google/univiz/api/representation/Deadline.java
+++ b/capstone/src/main/java/com/google/univiz/api/representation/Deadline.java
@@ -20,8 +20,6 @@ public abstract class Deadline {
 
   public abstract int closingYear();
 
-  public abstract String name();
-
   public static Builder builder() {
     return new AutoValue_Deadline.Builder();
   }
--- a/capstone/src/main/java/com/google/univiz/api/representation/Deadline.java
+++ b/capstone/src/main/java/com/google/univiz/api/representation/Deadline.java
@@ -7,6 +7,7 @@ import com.google.auto.value.AutoValue;
  * for the purpose of graphing due dates.
  */
 @AutoValue
+@GenerateTypeAdapter
 public abstract class Deadline {
   public abstract int openingMonth();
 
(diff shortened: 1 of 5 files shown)


capstone/src/main/java/com/google/univiz/api/representation/Deadline.java
```diff
@@ -7,4 +7,37 @@
  * for the purpose of graphing due dates.
  */
 @AutoValue
```
biancamacias, line 9:
```suggestion
@AutoValue
@GenerateTypeAdapter
```
and also add `import com.ryanharter.auto.value.gson.GenerateTypeAdapter;`

capstone/src/main/java/com/google/univiz/api/representation/Deadline.java
```diff
@@ -1,10 +1,45 @@
...
-public abstract class Deadline {}
+@GenerateTypeAdapter
+public abstract class Deadline {
+  public abstract int openingMonth();
+
+  public abstract int openingDay();
+
+  public abstract int openingYear();
```
ccondit, line 17: In general, it's good to leverage existing Date / Time libraries where possible.

Can you try modeling this as a pair of [LocalDate](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/time/LocalDate.html) instances? We'll also need to verify that GSON can serialize them...

capstone/src/main/java/com/google/univiz/api/representation/Deadline.java
```diff
@@ -1,10 +1,45 @@
...
+import com.ryanharter.auto.value.gson.GenerateTypeAdapter;
 
 /**
  * Contains the date when applications are due for a particular college. Will be made serializable
  * for the purpose of graphing due dates.
  */
 @AutoValue
-public abstract class Deadline {}
+@GenerateTypeAdapter
+public abstract class Deadline {
```
ccondit, line 12: Should there be a collegeId to tie it in to a particular college?

capstone/src/test/java/com/google/univiz/servlets/CollegeDataVisualizationServiceTest.java
```diff
@@ -81,9 +92,13 @@ public void getStatsResults_multiResult() throws Exception {
 
   @Test
   public void getDeadlineResults_throws() throws Exception {
```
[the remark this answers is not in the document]
ihsan314: I think I did this as you were reviewing.

capstone/src/main/java/com/google/univiz/api/representation/Deadline.java
ihsan314: I was planning on using an existing datatype for this, but was worried about GSON not being able to serialize it. Further I wasn't sure if it would deserialize into the [Date](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date) datatype or if it would just end up deserializing into something similar to the above.

All the same I'll try using LocalDate and check what happens.

capstone/src/main/java/com/google/univiz/api/representation/Deadline.java
ihsan314: The visualization will look something like this:
Heat Map

It will show how frequently and around what times applications are open/due, without attaching it to particular universities. That's why I didn't put the id or the name.

ccondit: SGTM

PR #110, opened 2020-08-24
Add graphs for demographical data

ihsan314:
Add graphs of these properties to Univiz:
- Admission rate
- Average SAT score
- Number of undergraduate students enrolled
- The ratio of male and female students

Here's what they look like:
image


ihsan314:
> Can you add a screenshot of the graphs on the `dashboard.html` to the PR description so that we know it shows up?

I've now edited the PR description to show the graphs.

capstone/src/main/webapp/js/demographicalGraphs.js
```diff
@@ -0,0 +1,184 @@
...
+async function deserializeNumUndergradsData() {
+  const data = new google.visualization.DataTable();
+  await fetch('/viz/stats')
+      .then((response) => response.json())
+      .then(
+          (numUndergradsInfo) =>
+            populateNumUndergradsDataTable(data, numUndergradsInfo),
+      );
+  return data;
+}
```
ccondit, line 96: This is going to result in a REST call for every graph. Is there a way we can consolidate those network calls to save resources?

I'm fine to address this in a follow up...

ihsan314: Ah yea I was planning on addressing this later. I think what I'll try is fetching and passing a parameter consisting of the JSON object. Then not only will network calls be reduced, but all the populate functions can run in parallel.

kaveh096: +1 to reducing network round trips.

PR #112, opened 2020-08-24
Displaying Map Feature

biancamacias:
This PR includes the front-end JavaScript code that displays the map feature and its main elements, which include markers where colleges are located and info windows with basic information (if we customize the info windows, we can add more to this). This PR includes the changes made to the `dashboard.html` that includes the restricted API key that displays the map. It also includes the `maps.js` file which includes the code that fetches the JSON maps data from the servlet and creates markers and info windows based on that data. 

The `mapsSpec.js` file has been mocked with as much `expect`s as I could add. Jasmine has been giving me a really hard time with mocking these `google.maps` objects, and also coupled with the fact that much of my code involves many anonymous functions, it was just really difficult to test the mock JSON data with the maps.js file. All functions work, however, as they have been tested by running `mvn package appengine:run` locally and also have been used in demos. 

Attached is a picture of some sample colleges, their markers, and their info windows, which ran locally. To test, I replaced `mapsDataList` from the `createMap()` function to be a hardcoded JSON array as you see in the test.
Screenshot 2020-08-24 at 11 02 51 AM


capstone/src/main/webapp/js/maps.js
```diff
@@ -0,0 +1,68 @@
...
+ */
+async function fetchData() {
+  // TODO(biancamacias): insert path to maps servlet
+  await fetch('/map')
+      .then((response) => response.json())
+      .then((mapsDataFromJSON) => {
+        console.log('Fetching data...');
+        console.log(mapsDataFromJSON);
+        return mapsDataFromJSON;
+      });
```
ihsan314, line 51:
```suggestion
      });
  return mapsDataFromJSON;
```

capstone/src/main/webapp/js/maps.js
```diff
@@ -0,0 +1,68 @@
...
+/**
+ * Helper function for setting content description for info windows.
+ * @param {string} name Name of college/university/institution
+ * @param {string} city City where college/university/institution is located in
+ * @param {boolean} isMainCampus True if it is the main campus, false otherwise
+ * @return {string} Description that will be used for info window content
+ */
+function setDescription(name, city, isMainCampus) {
+  const description = name + ' is located in ' + city + '. ';
+  if (isMainCampus == true) {
```
ihsan314, line 63:
```suggestion
  if (isMainCampus) {
```

capstone/src/main/webapp/js/maps.js
```diff
@@ -0,0 +1,68 @@
...
+
+    const marker = new google.maps.Marker({
+      position: latitudeLongitude,
+      map: map,
+      title: mapsData['name'],
+    });
+
+    (function(marker, mapsData) {
+      google.maps.event.addListener(marker, 'click',
+          function() {
```
ihsan314, line 27:
```suggestion
          () => {
```
Arrow functions are preferred according to the [style guide](https://google.github.io/styleguide/jsguide.html#features-classes-old-style).

capstone/src/main/webapp/js/maps.js
```diff
@@ -0,0 +1,68 @@
...
+        mapsData['latitude'],
+        mapsData['longitude']);
+
+    const marker = new google.maps.Marker({
+      position: latitudeLongitude,
+      map: map,
+      title: mapsData['name'],
+    });
+
+    (function(marker, mapsData) {
```
ihsan314, line 25:
```suggestion
    ((marker, mapsData) => {
```

capstone/src/test/javascript/mapsSpec.js
```diff
@@ -0,0 +1,22 @@
...
+          {'name': 'Stanford University',
+            'city': 'Stanford',
+            'isMainCampus': true,
+            'latitude': 37.429434,
+            'longitude': -122.167359},
+        ];
+        spyOn(window, 'fetch')
+            .and.returnValue(Promise.resolve({json: () => mapsJsonData}));
+        await fetchData();
+        // TODO(biancamacias): mock? assert functions work? but how?
```
ihsan314, line 20: So far I think the set up is pretty good. One testable aspect here is making sure that the constructors/methods are called properly (which would imply the Google Maps datatypes are being created/manipulated properly). Another would be any helper functions, like `setDescription`.

For the first aspect, you would spy on the constructor and the methods like this:
```javascript
spyOn(theClass.prototype, 'constructor');
spyOn(theClass.prototype, 'method');
// do something with theClass
expect(theClass.prototype.constructor)
  .toHaveBeenCalledWith(params);
expect(theClass.prototype.method)
  .toHaveBeenCalledWith(otherParams);
```

The second aspect should be more directly testable, as you won't need external functions, and can directly compare the actual result to the expected result:
```javascript
expect(helper()).toEqual(expectedResult());
```

If you want, we can discuss this further offline.

capstone/src/main/webapp/js/maps.js
biancamacias: i had this code before just like this, but i think javascript scope hinders me from doing this as i got an error saying it couldn't find `mapsDataFromJson` when i had the `return` outside the brackets. i haven't tested to see if it works with the servlet. i might end up just returning `response.json()` in the end for brevity, but im going to wait to address this.

ihsan314: Oh yea since `mapsDataFromJSON` is only defined within the last `then` block, it won't be accessible outside of it. Then yea it can be left alone.

biancamacias:
Currently failing tests, this is the error from `mapsSpec.js`. Any thoughts? Is my constructor not set up correctly?
image

ihsan314:
> Currently failing tests, this is the error from `mapsSpec.js`. Any thoughts? Is my constructor not set up correctly?
> image

You set up the mock correctly. But in the spy functions, you should specify `google.maps.prototype`, not `google.maps`.

biancamacias:
current error is below. i'm wondering if maybe `google.maps.InfoWindow.setContent` is the same instance as `google.maps.event`?
image

capstone/src/main/webapp/js/maps.js
```diff
@@ -0,0 +1,66 @@
...
+}
+
+/**
+ * Function that creates map markers and info windows.
+ * @param {JSON} mapsDataFromFetch JSON response from servlet
+ */
+function createMap(mapsDataFromFetch) {
+  const map = new google.maps.Map(
+      document.getElementById('map'),
+      {center: {lat: 39.8097343, lng: -98.5556199}, zoom: 5});
```
ccondit, line 25: Might be worth a comment for future readers about why this is the center and zoom level...

PR #115, opened 2020-08-26
Fetch network resources only once

ihsan314:
Currently in `demographicalGraphs.js` the network resources are fetched repeatedly, and the graphs are drawn synchronously. This PR fetches from the network only once and draws the graphs concurrently.

Visual:
image


ihsan314:
> nice cleanup! add a screenshot to make sure it displays?

Screenshot added.

PR #119, opened 2020-08-26
Integrate template from html5up.net/stellar

ihsan314:


ihsan314:
When reviewing, make sure to choose from the second commit onwards.

capstone/src/main/webapp/reference.html
```diff
@@ -3,21 +3,80 @@
...
+              <p>
+              During the course of creating UniViz, the following resources were used:
+              </p>
+              <ul>
+                <li>Stellar HTML5/CSS3 template from <a href="html5up.net/stellar">html5up.net</a></li>
+                <li><a href="https://collegescorecard.ed.gov/data/documentation/">The CollegeScoreCard API</a></li>
+                <li>The following Google APIs:
+                  <ul>
+                    <li>Google Charts API</li>
+                    <li>Google Places API</li>
```
biancamacias, line 28:
```suggestion
                    <li>Google Maps JavaScript API</li>
```
we ended up not using Places, so replace it with Google Maps JavaScript API

PR #120, opened 2020-08-27
Get ids from url before each fetch

ihsan314:


capstone/src/main/webapp/js/calendar.js
```diff
@@ -5,7 +5,9 @@
  */
 function drawCharts() {
   const frequencyData = new google.visualization.DataTable();
-  fetch('/viz/deadlines').
+  const queryURL = new URL(window.location.href);
```
ccondit, line 8: nit - camel case per style (and throughout this PR)
```suggestion
  const queryUrl = new URL(window.location.href);
```

PR #87, opened 2020-08-14
MapsResource.java class implementation

biancamacias:
This PR is about the implementation of the `getMapsData` method in the `MapsResource.java` class. This PR also includes `MapsResourceImplTest.java` class that tests the implementation. This PR also uses Guice to inject the `CollegeDataApi` interface to get a `List<CollegeData>` to convert to `List<MapsData>`.

PR #122, opened 2020-08-27
Filters College Names in java now

mmurray22:
Moves the filtering of college names from Javascript to Java. 

What does filtering mean in this context?
This means that the phrase that the user provides to the Scorecard API must be found (anywhere) in the word as the full phrase. So, when looking up 'New York U' you should get results containing the full phrase 'New York U' and not a result like '...University-New York...'


capstone/src/test/java/com/google/univiz/api/resource/SearchResourceImplTest.java
```diff
@@ -58,4 +63,16 @@ public void testNullCollegeName() throws Exception {
...
+  public void testFilteringIncorrectSuggestions() throws Exception {
+    String collegeName = "Stan";
+    when(mockSuggestionApi.getCollegeSuggestions(collegeName))
+        .thenReturn(MULTIPLE_SUGGESTION_RESPONSE);
+
+    List<SearchData> ret = search.getSearchSuggestions(collegeName);
+    CollegeId collegeId = CollegeId.create(STANFORD_SUGGESTION_DATA.id());
+    SearchData expected = SearchData.create("Stanford University", collegeId);
+    assertThat(ret).containsExactly(expected);
+  }
```
biancamacias, line 77:
```suggestion
  }
  
  @Test
  public void testIoException() {}
```
i would add a test to make sure it accepts the exception. (chris would probably ask for a test to make sure they throw exceptions haha). The `MapsResourceImplTest.java` class in #87 has an example of it.

mmurray22: done!

mmurray22:
> This probably still has the same bug as Kaveh mentioned in yesterday's standup, where the suggestions empty out midway through the string `"University"`. But for now I guess it works, if we assume the user is always going to enter a string that corresponds to the first few letters of the name of the college they have in mind. Beyond that, I think it's good.

Yes it does :) The solution to the bug you are mentioning is in a chained PR I plan to release later today :)

ccondit:
1. Can you link this PR to an issue?
2. This actually seems like a feature we're removing from the application. Wouldn't it save users time to be able to autocomplete on words towards the end of the name of the institutions? For instance if I'm searching for `Berk` in hopes of finding `University of California - Berkeley`, I now have to:
    - Have knowledge of the precise name
    - Spend the time to type out that potentially long name

kaveh096:
I'm still not clear why we need to filter these results? What's bad about showing suggestion results that do not start with the query? So let's sync tomorrow about this. Also whether we want to actually do this filter or not, let's priotize first getting the auto-complete feature running end to end before adding extra features (like filtering and paging) to it.

PR #123, opened 2020-08-27
Gets all possible suggestions

mmurray22:
Gets all suggestions from all pages provided by the College Scorecard API and picks the 20 most relevant to the user's query.

NOTE: Currently, it seems the JS is outdated given the new Java implementation that is definitely taking on a much higher load in terms of data. Any suggestions to make this better would be appreciated.


mmurray22:
> I think now this PR takes care of the "University" bug, but what about what Chris mentioned in [this comment](https://github.com/googleinterns/univiz/pull/122#issuecomment-682079777) regarding searching at the end of the word? Another problem I think could be acronyms, like searching `ASU` in the hopes of finding either `Alabama State University` or `Arizona State University`. I understand that we aren't going to be implementing PageRank or anything fancy, but I think we should at least see if we can do something in addition to checking the beginning of the string.
> 
> An idea I have is if in the REST call you sort the results by descending student population, you can get the top `MAX_SUGGESTIONS` schools by student body size that match the given query. This could mess with a lot of other things though, so maybe revisit this suggestion if nothing else works.
> 
> Otherwise, the addition of the `MetaInfo` and `MetadataReader` AutoValues are pretty nice. And handling the results from the Java side rather than the JavaScript side allows for more effective solutions.

Thanks for the input Ihsan :) I think my biggest aversion to searching for colleges which have that substring anywhere in their name is that the general expected functionality of this search engine is that a user starts typing in the beginning of the college's name and then suggestions based on that pop up. It seems a bit unintuitive for the suggestion 'Stanford University' to show up with the user types in 'University', at least to me. Also, in most sites that allow you to specifically only search up colleges, it also seems to model this behaviour. Google ofc can give you tuned suggestions no matter what substring of the word you are suggesting, but I was thinking this more straightforward behaviour would be more readable for the user.

PR #129, opened 2020-08-31
Adds the remove feature to the autocomplete interface.

mmurray22:
Currently, if you select a college, there is no way to de-select said college. This PR aims to solve this smoother functionality hinderance.

Screenshot:
remove-function


capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -88,6 +88,32 @@ function closeAllElements() {
...
+ */
+function keepTrackOfChosenColleges(validSuggestion) {
+  const parent = document.getElementById('suggestions');
+  const listElement = document.createElement('li');
+  listElement.setAttribute('id', validSuggestion.collegeId.id);
+  listElement.innerHTML = validSuggestion.collegeName;
+  listElement.innerHTML += ' ';
+  const removeButton =
+    '<button onclick=\'removeCollege('+
+    validSuggestion.collegeId.id + ')\'>X</button';
```
ihsan314, line 103: The X may not render as nicely as we want, since it won't show up like ❌ or ✖️ or anything like that. It will just show up as text. I would suggest either using a special Unicode emolji/character, or just use the string "Remove" instead of "X".

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -88,6 +88,32 @@ function closeAllElements() {
...
+  listElement.innerHTML += removeButton;
```
ihsan314, line 104: Consider using the DOM to create the button:
```javascript
const removeButton = document.createElement('button');
removeButton.onClick = () => removeCollege(validSuggestion.collegeId.id);
removeButton.innerText = 'X' // or the replacement you choose
listElement.append(removeButton);
```

capstone/src/main/webapp/js/autocomplete.js
ihsan314: image
This is how it renders right now. I think in addition to replacing the "X", the button should appear below the university name.

mmurray22: done :)

capstone/src/main/webapp/js/autocomplete.js
mmurray22: updated :)

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -118,6 +127,11 @@ function keepTrackOfChosenColleges(validSuggestion) {
   const listElement = document.createElement('li');
   listElement.setAttribute('id', validSuggestion.collegeId.id);
   listElement.innerHTML = validSuggestion.collegeName;
+  listElement.innerHTML += ' ';
```
ihsan314, line 130:
```suggestion
  listElement.innerHTML += '<br> ';
```
so that it looks like this:
image

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -107,8 +116,8 @@ function createAutocompleteListElement(collegeName, value) {
 }
 
 /**
- * Adds valid suggestions to list of stored suggestions
- * @param {string} validSuggestion
+ * Keeps track of chosen colleges
+ * @param {Array<string, collegeId>} validSuggestion
  */
 function keepTrackOfChosenColleges(validSuggestion) {
```
ccondit, line 122: Can you add a simple test verifying that this method works? No need to test removes, just that this method is roughly WAI...

mmurray22: will do! give me five minutes @ccondit !

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -118,6 +127,11 @@ function keepTrackOfChosenColleges(validSuggestion) {
   const listElement = document.createElement('li');
   listElement.setAttribute('id', validSuggestion.collegeId.id);
   listElement.innerHTML = validSuggestion.collegeName;
+  listElement.innerHTML += ' ';
+  const removeButton = document.createElement('button');
+  removeButton.onclick = () => removeCollege(validSuggestion.collegeId.id);
```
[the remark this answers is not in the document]
ccondit: Whoops - I didn't mean to actually commit this...

capstone/src/main/webapp/js/autocomplete.js
mmurray22: haha, well that wasn't five minutes, but done!

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -120,18 +120,20 @@ function createAutocompleteListElement(collegeName, value) {
 }
 
 /**
- * Adds valid suggestions to list of stored suggestions
- * @param {string} validSuggestion
+ * Keeps track of chosen colleges and the return value is for testing
```
ccondit, line 123: nit - you've refactored this s.t. the return value is used in code. update the comment to reflect this.

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -120,18 +120,20 @@ function createAutocompleteListElement(collegeName, value) {
...
+ * @param {Array<string, collegeId>} validSuggestion
+ * @return {HTMLDivElement} listElement
  */
 function keepTrackOfChosenColleges(validSuggestion) {
```
ccondit, line 127: Now that this function has been refactored, can you update the function name to reflect what it does?

capstone/src/main/webapp/js/autocomplete.js
mmurray22: done!

ccondit (approved):

capstone/src/main/webapp/js/autocomplete.js
```diff
@@ -147,7 +149,13 @@ function displayCollegeSuggestions(suggestions, autocompleteList, value) {
     const copyArrayElement = arrayElement;
     listElement.addEventListener('click', () => {
       SEARCH_INPUT.value = copyArrayElement;
-      keepTrackOfChosenColleges(copyArrayElement);
+      if (document.getElementById(copyArrayElement.collegeId.id)) {
```
kaveh096, line 152: I don't understand why we need this `if`. Are you worried that the click event might get triggered after the list element is removed?

kaveh096 (approved):

capstone/src/main/webapp/js/autocomplete.js
mmurray22: no, it's to address the instance where 'Stanford University' has already been selected and the user tries to reselect 'Stanford University'

